Issue #107, opened 2020-03-12
Meeting Today?


jonchurch:
A tracking issue for a meeting was never created. Are we meeting today?

gireeshpunathil:
I asked it here: https://github.com/expressjs/discussions/issues/100#issuecomment-598000370 I was under the impression that you create those manually.

Either way, @expressjs/express-tc @expressjs/triagers - can we make it in the next 30 minutes? I am ! 

Please indicate here through thumbs up!

jonchurch:
I created the last two because no one else did but didnt get to it this time.

gireeshpunathil:
haven't heard from @dougwilson or @wesleytodd who have the handles to run the TC, so let us deem it cancelled for today!

@expressjs/express-tc - please post your next availability!

ryhinchey:
I’m generally available just need to know ahead of time!

wesleytodd:
Was dealing with changing travel plans so wouldn’t have been able to make it anyway. I think our plan is in action, was there anything specific we needed to talk about anyway?

gireeshpunathil:
> I think our plan is in action, was there anything specific we needed to talk about anyway?

@wesleytodd - If it makes sense to others, I would love to have a working session in the next TC, basically targetting PRs that are low hanging, or having minor differences to ratify. I guess it makes a lot easier if people do it together and land those, especially the ones that are easy ones, ready to land and are ear-marked for 5.

what do you think? If people like it, I could identify a few that we can work it out in the TC. Let me know!

jonchurch:
I do think we are in a good place w/ v5, so meeting today wasn't necessary anyways.

Leaving some notes based on reducing confusion in the future. I've experienced these things in other open source projects as well, and think it's a good idea for project health to address them eventually.

**Takeaways:**

* A meeting was assumed to happen today because we said two meetings ago that we were going to do weekly meetings to check in on v5
* I don't believe we officially documented that decision, or the schedule, outside of issue comments or a recording
* There wasn't a pending issue to track the next meeting, so folks (like myself) weren't sure if it was official. 
* Without key stakeholders (permissions for zoom, streaming), it's unclear how to proceed with an ad hoc meeting if some folks want to hold one

**Remedies:**

* Make sure meeting cadence is committed to the readme, update it when cadence changes (e.g. when we put forward doing weekly)
[...]

Issue #103, opened 2020-02-26
Express TC Meeting 02-26-2020 (ad hoc last minute)


jonchurch:
## Attendance

The entire community is welcome to tune in and observe the meeting live at the time below. The following people are listed here as explicit invitees to get an email notification about the upcoming meeting:

Invited:

@expressjs/express-tc
@expressjs/triagers

## When
Feb 26, 2020 - 23:30 UTC

## Where
**Watch the recording:** https://www.youtube.com/watch?v=_GRZKdu3bKY
Zoom https://zoom.us/j/884354313

* Viewers: https://www.youtube.com/watch?v=_GRZKdu3bKY
* Participants: @dougwilson @wesleytodd @gireeshpunathil @LinusU @HarshithaKP @jonchurch @bethgriggs @mlrawlings 

## Agenda

Request for TC Meeting to talk about 5.0 #99
Express 5.0.0 #100 
Release 5.0 https://github.com/expressjs/express/pull/2237 (Specifically https://github.com/expressjs/express/pull/2237#issuecomment-570058548)
Faster Major Release Schedule https://github.com/expressjs/discussions/issues/67
Document and make official repo captains #98
Express YouTube Channel #94
Express triage #92

wesleytodd:
Notes:

- We decided on 4:00 pm GMT on Thursday March 5th for the next meeting
  - @wesleytodd to get it on the calendar
  - @dougwilson and @wesleytodd  to figure out a better streaming situation before then
- The [list of todo items](https://github.com/expressjs/express/pull/2237#issuecomment-570058548) looks good in general
- Everyone should review the repo capitans proposal as it will be our first topic next week
  - https://github.com/expressjs/discussions/issues/98
- Second agenda item for next week is to follow up on the plan based and divide the work

jonchurch:
@dougwilson ping, can you please update the issue with a link to the recording on youtube once it's available?

dougwilson:
@jonchurch apologies, I left it while it was processing. Linked added above 👍

gireeshpunathil:
resolved, closing

Issue #99, opened 2020-02-23
request for a TC meeting to discuss express 5 release


gireeshpunathil:
as per https://github.com/expressjs/express/issues/4195#issuecomment-590057521

please include this in the agenda:
 - user feedback so far on the alpha releases
 - decide whether `5.0.0-alpha.7` can be cut into express 5.0.0
 - critical blockers if any
 - milestones, owners, checkpoints, commitments, availability matrix

dougwilson:
@expressjs/express-tc and OP, do we think this Wednesday Feb 26, 2020 - 23:30 UTC would be good, or do we need until next week for any conflicts?

wesleytodd:
This time works for me!  

Sorry everyone, with my new position at work I have been slammed.  I intended to follow up better on some of this stuff, but paid work comes first!  I posted https://github.com/expressjs/express/pull/2237#issuecomment-570058548 on Jan 1 with optimism :).  But in general I think we need to have a team agreement on what to do with those remaining items on the list.  A TC meeting for this would be great!

I also propose that we ensure that we have enough time so that we reach final decisions and don't let it hang longer if possible.  It would be awesome if as many project participants can come as possible, so we should also tag @expressjs/triagers.

dougwilson:
I would suggest that we even have a TC meeting starting this Wednesday and again every Wednesday until we actually have the release complete, at a minimum. I feel like doing this will go a long way towards preventing momentum from stalling out.

gireeshpunathil:
> do we think this Wednesday Feb 26, 2020 - 23:30 UTC would be good

It is 5 AM (IST) for me, but will make it if others are good with the time.

LinusU:
> do we think this Wednesday Feb 26, 2020 - 23:30 UTC would be good

It's 00.30 for me, but I can also make it work if others are good with the time 👍

jonchurch:
Even if we don't discuss Triage, I would like to attend the meeting to observe, so count me in. 

I believe that time is 6:30PM EST Feb 26th

gireeshpunathil:
@jonchurch - yes, that looks right.

so looks like we are all good for 23.30 UTC! @dougwilson / @wesleytodd - will you pls confirm the meeting and create and share a link?

dougwilson:
Awesome, everyone! The time is definately open to change; it was decided at the time we had a person in Australia, but that is no longer the case. Once we see who is going to be attending, I think we can easily adjust the time to better fit the attendies.

[...]

Issue #95, opened 2019-12-06
Restructure TC Meeting With Triagers


wesleytodd:
As came up in the triage sync meeting is that we will want some time set aside for discussion triage topics.  As the goal of this is to have these new contributors participate as much as possible in the project, I think it would be reasonable to restructure the TC meeting and schedule to accommodate an additional 30min to discuss triage topics.

It would be great if we could make a decision and set a time for early-mid January to kick off the changes.  That way we could promote it at Node Interactive and get a good group of people together for the first meeting.  If we are not ready to change the whole TC meeting then we should at least set a time for the triage sync.

For reference: https://www.youtube.com/watch?v=Y0nqjseQbN0

cc: @expressjs/express-tc @expressjs/triage

wesleytodd:
One other thing we might consider is changing the name to "Community Meeting"?  I think one of the goals is to make sure everyone feels like they can speak up and be included.  If they are not "on the tc" it might make people feel less comfortable giving input and feedback.  Maybe this is not a real issue, but wanted to at least bring it up.

dougwilson:
As for the name, I think it just came from following the Node.js setup; the actual name doesn't matter at least to me :D 

Who should be in charge of the decision above? The length of the meeting has always been wishy-washy and so adding 30 mins at the end would just be a function of the invite. Setting a new time and day of the week, ok the other hand, should just be a doodle. I know last I remember there was going to be a triage meeting early Jan. Was this discussed at that time at all?

Just trying to understand what landed here in the interim of time :)

wesleytodd:
I think we didn't finalize any meetings changes.  If we are going to try and have a TC meeting this week or next then I think we should discuss this so we can move forward on it.

jonchurch:
I'm happy to attend the next meeting and chat about Triage or just observe 👍

ryhinchey:
I'm not officially a triager yet but have been starting to help out.  I'm happy to attend a meeting about triage

wesleytodd:
@ryhinchey, I am sorry for the hold up, life has been really busy lately, and realizing that I had multiple groups to manage I started to write automation for this instead of just doing it manually.  My engineer brain led me to think automating it was a better use of my time lol, what was I thinking 😛

ryhinchey:
no worries @wesleytodd! life happens

gireeshpunathil:
we are already practicing this, so just wondering if there is anything left here:
 - just close
 - document the current process and close

[...]

PR #108, opened 2020-03-12
Update readme with v5 meeting cadence, and ad hoc meeting info

jonchurch:
In reference to https://github.com/expressjs/discussions/issues/107#issuecomment-598300865

Please suggest any edits inline, wanted to make sure I opened a PR since I brought up these points.

> **Takeaways:**
> 
> * A meeting was assumed to happen today because we said two meetings ago that we were going to do weekly meetings to check in on v5
> * I don't believe we officially documented that decision, or the schedule, outside of issue comments or a recording
> * There wasn't a pending issue to track the next meeting, so folks (like myself) weren't sure if it was official. 
> * Without key stakeholders (permissions for zoom, streaming), it's unclear how to proceed with an ad hoc meeting if some folks want to hold one
> 
> **Remedies:**
> 
> * Make sure meeting cadence is committed to the readme, update it when cadence changes (e.g. when we put forward doing weekly)
> * Make sure issues are generated to track a meeting. I'm in favor of anyone being able to create them (since they are easily editable/close-able), and suggest it is part of the wrap up of any given meeting. Issues are generated automatically w/ the Node.js projects based on a calendar and [some templates](https://github.com/nodejs/create-node-meeting-artifacts), which we could set up eventually, but let's not overengineer things too early.
> * Document procedure for holding ad hoc meetings, or in the case that the main stakeholders aren't available. Can just be `If stakeholders aren't available, start a google hangout, post the link, and have someone record it`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,20 @@
 
 This repository is used for generalized Express.js discussions.
 
-* [Current discussions](https://github.com/expressjs/discussions/issues?q=is%3Aissue+is%3Aopen+label%3Adiscuss)
-* [Upcoming meetings](https://github.com/expressjs/discussions/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3Ameeting)
+- [Current discussions](https://github.com/expressjs/discussions/issues?q=is%3Aissue+is%3Aopen+label%3Adiscuss)
+- [Upcoming meetings](https://github.com/expressjs/discussions/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3Ameeting)
+
+## Current v5 Meeting Cadence
+
+While focusing on releasing Express v5, the meeting schedule has been changed from the normal schedule listed below. Until further notice, meetings will be held on this cadence.
+
+**Currently, meetings are held weekly on Thursdays at 4pm GMT.**
+
+### If Stakeholders Not Available
+
+If for some reason the host of the meeting is not available, but there are participants who wish to discuss the agenda or other topics, an ad hoc meeting can be started. If the host is not available for a scheduled meeting, participants can post a link to a video chat in the scheduled meeting's issue.
+
+Meetings should be recorded when appropriate, like regular meetings are, and provided to the TC for uploading to the official youtube channel.
 
 ## Express.js TC meetings
 
@@ -12,4 +24,4 @@ a live stream of the discussion and a recording of past meetings. Topics for the
 meetings are determined on an as-needed basis and may be technical or management
 of nature.
 
-Current meeting schedule: Wednesdays 23:30 UTC
+Normal meeting schedule: Wednesdays 23:30 UTC
--- a/README.md
+++ b/README.md
@@ -2,20 +2,26 @@
 
 This repository is used for generalized Express.js discussions.
 
-- [Current discussions](https://github.com/expressjs/discussions/issues?q=is%3Aissue+is%3Aopen+label%3Adiscuss)
-- [Upcoming meetings](https://github.com/expressjs/discussions/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3Ameeting)
+* [Current discussions](https://github.com/expressjs/discussions/issues?q=is%3Aissue+is%3Aopen+label%3Adiscuss)
+* [Upcoming meetings](https://github.com/expressjs/discussions/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3Ameeting)
 
 ## Current v5 Meeting Cadence
 
-While focusing on releasing Express v5, the meeting schedule has been changed from the normal schedule listed below. Until further notice, meetings will be held on this cadence.
+While focusing on releasing Express v5, the meeting schedule has been changed
+from the normal schedule listed below. Until further notice, meetings will be
+held on this cadence.
 
 **Currently, meetings are held weekly on Thursdays at 4pm GMT.**
 
 ### If Stakeholders Not Available
 
-If for some reason the host of the meeting is not available, but there are participants who wish to discuss the agenda or other topics, an ad hoc meeting can be started. If the host is not available for a scheduled meeting, participants can post a link to a video chat in the scheduled meeting's issue.
+If for some reason the host of the meeting is not available, but there are
+participants who wish to discuss the agenda or other topics, an ad hoc meeting
+can be started. If the host is not available for a scheduled meeting,
+participants can post a link to a video chat in the scheduled meeting's issue.
 
-Meetings should be recorded when appropriate, like regular meetings are, and provided to the TC for uploading to the official youtube channel.
+Meetings should be recorded when appropriate, like regular meetings are, and
+provided to the TC for uploading to the official youtube channel.
 
 ## Express.js TC meetings
 


README.md
```diff
@@ -2,8 +2,20 @@
 
 This repository is used for generalized Express.js discussions.
 
-* [Current discussions](https://github.com/expressjs/discussions/issues?q=is%3Aissue+is%3Aopen+label%3Adiscuss)
-* [Upcoming meetings](https://github.com/expressjs/discussions/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3Ameeting)
+- [Current discussions](https://github.com/expressjs/discussions/issues?q=is%3Aissue+is%3Aopen+label%3Adiscuss)
+- [Upcoming meetings](https://github.com/expressjs/discussions/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3Ameeting)
```
dougwilson: Why did these need to be changed?

README.md
```diff
@@ -2,8 +2,20 @@
...
+
+## Current v5 Meeting Cadence
+
+While focusing on releasing Express v5, the meeting schedule has been changed from the normal schedule listed below. Until further notice, meetings will be held on this cadence.
```
dougwilson: This (and the other lines) should be hard wrapped like the rest of the doc :)

README.md
```diff
@@ -12,4 +24,4 @@ a live stream of the discussion and a recording of past meetings. Topics for the
 meetings are determined on an as-needed basis and may be technical or management
 of nature.
 
-Current meeting schedule: Wednesdays 23:30 UTC
+Normal meeting schedule: Wednesdays 23:30 UTC
```
dougwilson: Rather than change this and add need to add the further notice warnings, etc. What if (especially if they are going to end up as "ac-hoc" meetings) we just don't call the Thursdays meetings Express.js TC Meetings? Maybe like Express.js Roadmap Meetings? This is not a request to change anything if you don't want here, just a thought I had regarding it. I do kind of feel bad that, if this is the TC meeting, having it moved where a key member (myself) cannot even commit to being there due to conflicting with work. I just cannot jeopardize my job that puts my roof over my head for something that does not, if that makes sense, haha

gireeshpunathil: @dougwilson  - on the other hand, express 5 being important to the ecosystem, the community and everyone involved here, it is critical that express TC is committed to its execution plan from start to end, driving from the front. For that reason, I would maintain that we call it and run it as Express TC meetings; not ad-hoc, and make any adjustments possible to accommodate key persons like you.

dougwilson: I agree it is of up-most important and 100% want to be at the meetings... but I'm not sure if the meeting time/day is up for reconsideration currently as I have brought up the conflict before and the meeting remains at a conflicting time, haha

dougwilson: My thought was if this is the "TC Meeting" then should the schedule be set to try and accommodate all the TC members at most and others can attend if able who are not in the TC? If the meeting is set to accommodate more than the TC members alone, that feels like it is no longer a TC meeting, but a more general, wider, Express.js meeting (which I have nothing against, of course).

gireeshpunathil: @dougwilson - no other TC members than you, have indicated any inconvenience to the current time, IIUC.

The meeting is never set to accommodate non-TC members alone, I don't know how you got that impression?

dougwilson: AFAIK the existing TC meeting didn't have any issues with the current TC members. What was the reason for changing the meeting time/day in the first place? Why not go back to the time in the README?

dougwilson: > The meeting is never set to accommodate non-TC members alone, I don't know how you got that impression?

My impression came from #95 , unless I am completely misunderstanding that issue?

gireeshpunathil: > What was the reason for changing the meeting time/day in the first place? Why not go back to the time in the README?

@dougwilson - answer in your own comment
https://github.com/expressjs/discussions/issues/99#issuecomment-591204805

dougwilson: Yea, I am open to it changing. But why was it changed to when I can't actually attend, lol

gireeshpunathil: > My impression came from #95 , unless I am completely misunderstanding that issue?

right - but that issue does not propose to accommodate non-TC at the cost of keeping TC away. May be I misunderstood your original comment as hinting on these lines, but reading it again, I don't think it was put that way.

Let us all together uphold the value of express TC, the weight of discussions happened there, and be accountable for decisions taken in the meetings - for the interest of the community and the users that depend on it.

dougwilson: Yep, I agree. Does this mean we can move the TC meeting back to the original time, or am I just never going to be able to attend a TC meeting for the foreseeable future :) ?

dougwilson: > Let us all together uphold the value of express TC, the weight of discussions happened there, and be accountable for decisions taken in the meetings - for the interest of the community and the users that depend on it.

Also, I'm not sure what exactly that means, as it almost sounds like if a decision is reached in a meeting is somehow has more weight than otherwise, which is not what our currently process is. The process is outlined inhttps://github.com/expressjs/express/blob/master/Contributing.md#tc-process which states "The group tries to find a resolution that has no open objections among TC members."; a TC member can have an objection to something even if they are not present in a TC meeting, as otherwise it would not make sense to hold one without all members present.

There is a process for when a TC member as an open objection and a consensus cannot be reached, which is through the majority vote method (has never been used before). There are currently 5 active TC members so only 3 votes would be necessary for a majority, if we actually have to do such a thing.

gireeshpunathil: are you talking about the timing decisions? I don't think we made it as a `TC decision`, but a consensus among members present.

So how about a doodle method? if that does not work out well, let us consider invoking the voting method as well.

dougwilson: I'm referring to your specific statement -- "accountable for decisions taken in the meetings"; that goes against the current governance, as a decision cannot be final inside a meeting unless there is actually consensus among all the TC members. It also means that if not all members are present for a meeting, by nature a consensus cannot be reached during a meeting...

gireeshpunathil: ok, that statement cannot be cut in the middle. It is prepended with `value of express TC` and appended with `for the interest of the community` , and has to be read together, as is.

> a TC member can have an objection to something even if they are not present in a TC meeting, as otherwise it would not make sense to hold one without all members present.

> a decision cannot be final inside a meeting unless there is actually consensus among all the TC members. It also means that if not all members are present for a meeting, by nature a consensus cannot be reached during a meeting...

these are not present in the clauses, and are just extrapolations. Strictly speaking, incorrect.

dougwilson: Yes, they are an extrapolation, but are, strictly speaking, 100% correct. You see, if something is decided in a meeting without all the TC members present, then those members do not have the opportunity to object to it at that time. So, they are 100% allowed to object after the meeting took place and they have listened to it / read the notes on it. Once there is an objection, there is not consensus. But during the meeting without all members present, it is _unknown_ if there is consensus at that point.

README.md
jonchurch: Linter auto-change I didnt notice

README.md
gireeshpunathil: > You see, if something is decided in a meeting without all the TC members present, then those members do not have the opportunity to object to it at that time.

I understand the sentiment, but the clauses do not mention anything about absentees and their rights; which makes sense too. Otherwise no meaningful discussion and consensus is possible in a meeting if a member is missing.

> Once there is an objection, there is not consensus

makes sense, but `objection` should be better qualified and scopped -  within the meeting, while consensus seeking process is on. Decisions cannot be postponed indefinitely, in expectation of future objections?

dougwilson: I'm not saying that meetings cannot try and make decisions. Only that just because they tried does not mean that a TC member who was or was not there (one who was there can change their mind of course) can later object. And you cannot dismiss their objection simply because they were unable to attend the meeting, especially if they expressed being unable to before hand and there wasn't an opportunity to take part in the discussion before-hand in an async fashion like over Github.

Just as you state that the clauses do not mention anything about absentees and their rights, it too does not state anywhere that decisions made in a meeting carry any more weight than done anywhere else, nor does it say that they are final in a meeting, etc.

All discussions are meaningful, no matter what medium they occur in.

> makes sense, but objection should be better qualified and scopped - within the meeting, while consensus seeking process is on. Decisions cannot be postponed indefinitely, in expectation of future objections?

I'm not sure what this mean. Again, you seem to be arguing that some specific things are not explicitly in the clause, and this, too, is not specifically in the cause. There is no definition for when a consensus seeking process starts or stops. The closest you have is that it calls out specifically about objections existing. 

> Decisions cannot be postponed indefinitely, in expectation of future objections?

I have never said that, either. Once something is spoken of, it is reasonable to proceed as if there is a presumed consensus. If an object does happen to appear in the future, and whatever the thing is hasn't actually be done, then it's clearly no longer a consensus on something that hasn't even happened yet.

If the ideal state you would like to see is to reduce the chance of objections post-meeting, it seems there are some pretty easy paths there (this is not comprehensive): you can attempt to schedule the meeting when those likely to object can attend, or you can open a discussion over some async meeting like Github  prior to the meeting so objections can be worked through. Working through objects can occur over any medium, including meetings themselves.

One of the main purposes of recording meetings is in order to allow the TC members to reach consensus by giving the opportunity to those who were unable to take part to understand what is happening and raise objections. This is similar to the requirement for pull requests to wait for a period of time, even when reaching the needed number of approvals, such that one has an opportunity to raise an objection, if they wish.

I would never argue that any decision can just be objected to with indefinite time, as that is arguing a nonsensical extreme. But if a TC member raises an objection within the same time period as a PR has to wait (36 hours) after the meeting's recording is uploaded for listening, I think that is a very reasonable amount of time, do you disagree?

gireeshpunathil: @dougwilson - +1 to almost everything you said, except:

> you can attempt to schedule the meeting when those likely to object can attend

may be this is impractical to achieve? how do one know in advance that someone is likely to object?

And moreover, since we are riding on a path that require high level of engagement and commitment leading up to a series of decision making, the natural expectation is that people are positive, progressive and accommodative by default.

Let us move on! I have seen you set up a TC at the regular schedule. It is 5 AM for me, but will surely accommodate - really excited to take part in moving this forward!

dougwilson: > may be this is impractical to achieve? how do one know in advance that someone is likely to object?

Correct, it is not really possible to _know_ who will object. That is why I specified "those who are likely to object" -- perhaps one way think that based on prior conversations / etc. I think it would be unreasonable to actually be able to know that, of course. Just as it is reasonable to say you don't know who would object, it is reasonable to give some small window for objection post-meeting is all.

> the natural expectation is that people are positive, progressive and accommodative by default.

I 100% agree, and unless there is something else going on here (perhaps through tone projection onto text, etc.)? I'm not aware of anyone acting in a way trying to _prevent_ progressing movement. Someone disagreeing on a direction, especially with an alternative direction presented, is by nature acting in good faith. Throwing up an objection without providing a reason would be obstructing; if the person who objects does provide a reason, but it isn't clear, it is fair to ask them to better clarify on it and provide specific questions to help drive the clarity for both sides to better understand each other -- they may actually be in agreement in the end, but saying the same thing in two different ways.

jonchurch: It sounds to me like we need to revisit the scheduling of the v5 meetings. 

From what I recall, we put forward meeting at the time outlined here after the first ad hoc meeting #103 based on the preferences expressed by those in attendance. If that time does not work for TC members and main stakeholders, then it makes sense to try and get consensus on another time. 

We want/need @dougwilson to be able to attend in order for meaningful progress to be possible. I hope we can also accommodate colleagues in timezones in the eastern hemisphere, and @gireeshpunathil your flexibility is greatly appreciated. 

If someone can start a doodle or equivalent to try and find a suitable meeting time to settle on, that would be awesome. I'm on the East coast of the US and very flexible personally, so I'll defer to the preferences of those we wish to accommodate. I don't want this to be a blocker for progress, and again thank you @gireeshpunathil for being so flexible on this issue currently, but I would be happy if we find a time that is palatable to all parties.

README.md
jonchurch: Updated wrapping to 80 chars

README.md
jonchurch: These changes have been undone

jonchurch:
Closing this because I think the discussion it pulled forth has concluded, and the proposed changes are no longer relevant.

Issue #196, opened 2024-02-24
EFI: Express LTS Strategy


wesleytodd:
Many decisions on which tools to use, coding patterns to follow, and generally run some technical decisions depend on how we decide to treat Node.js Version support and how to run major version releases. There is one requirement for an LTS strategy for this project:

1. We must maintain strict Node.js version support within a single `express` major

There are some really nice to haves:

1. Keeping our version support modern so we don't block node core from making progress
2. Keeping our version support modern so we don't get old outdated bug reports or security issues
3. Keeping our version support modern so we can adopt newer patterns faster

And then there are a lot of things to discuss around the cost and priorities of these things. I think we need to have this discussion, and I was waiting on things getting sorted more generally, but to keep the conversations focused I am opening this issue now so folks can centralize the conversation someplace without rehasing it in many different dependent issues like #184.

ljharb:
> Keeping our version support modern so we don't block node core from making progress

This merely requires continually testing on all supported versions of node, which both myself and pkgjs have actions to help you dynamically create a matrix for.

> Keeping our version support modern so we don't get old outdated bug reports or security issues

I'm not sure what this means - as long as modern node is supported, a bug on an older supported version of node is still a bug.

> Keeping our version support modern so we can adopt newer patterns faster

this really only applies to syntax, and that can be transpiled, so it needn't be a forcing function most of the time.

------------

One of the inherent negative costs of *any* major bump is that people will be left behind on the old major, and backports for bug and security fixes become much more annoying and unlikely. When a major bump drops support for an engine, it makes upgrade the platform for that user much harder. I strongly suggest not arbitrarily adhering to someone else's support plan (LTS, eg) and instead, decide which node versions to support based on what minimal featureset is needed for express to support the user patterns it wants, and no fewer.

In the event that a dependency has a tighter contraint than express otherwise needs to impose, the tradeoffs should be ad-hoc weighed between letting that dependency make decisions for this project, vs the cost of finding an alternative dependency. (in most cases, i'll be happy to maintain an alternative dependency if that resolves this question)

kibertoad:
> this really only applies to syntax, and that can be transpiled

Transpilation is not the best approach for a framework, as that can cause unpredictable runtime overhead in some cases, especially when transpiling to a much older version. And it complicates debugging significantly.

kibertoad:
@ljharb If the approach is to base supported version on desired libraries / syntax, it might make sense to start forming a wishlist for these. I'll start:

[...]

PR #199, opened 2024-02-26
Express LTS strategy

kibertoad:
refs #196

### DIFF
--- a/docs/LTS-strategy.md
+++ b/docs/LTS-strategy.md
@@ -2,16 +2,16 @@
 
 ## Goals
 
-* Maintain strict Node.js version support within a single Express major release.
-* Provide good ergonomics and developer experience for Express maintainers
+* Maintain strict Node.js version support within a single Express major release
 * Provide good experience for users who want to keep their Express projects that up-to-date
 * Avoid stagnation and blocking innovation that is good for the end users
+* Ensure good ergonomics and developer experience for Express maintainers
 
 ## Non-Goals
 
 * Only support the active LTS releases of Node.js
-* Support the widest possible set of Node.js versions. We aim to maintain a healthy balance between backwards compatibility and toolchain modernization.
 * Provide timed cadence of Node.js version support dismissal
+* Support an extended set of Node.js versions to the detriment of all other factors. We aim to maintain a healthy balance between backwards compatibility and toolchain modernization
 
 ### Rationale behind dropping support for old Node.js versions
 
--- a/docs/LTS-strategy.md
+++ b/docs/LTS-strategy.md
@@ -10,7 +10,7 @@
 ## Non-Goals
 
 * Only support the active LTS releases of Node.js
-* Provide timed cadence of Node.js version support dismissal
+* Release semver major versions on a regular cadence just for the sake of dropping Node.js version support when there are no clear benefits for doing so, and there are no other breaking changes planned for the release.
 * Support an extended set of Node.js versions to the detriment of all other factors. We aim to maintain a healthy balance between backwards compatibility and toolchain modernization
 
 ### Rationale behind dropping support for old Node.js versions
@@ -36,3 +36,7 @@
 * No features are backported to older semver major branches
 * Critical bugfixes will be backported to older supported semver major branches
 * Security fixes will be backported to older supported semver major branches
+
+### Release calendar
+
+* There is a regular cadence of semver major releases (details TBD). However, if no breaking changes were introduced between the last semver major and the scheduled date of the release of the new one, the new semver major release is skipped.
--- a/docs/LTS-strategy.md
+++ b/docs/LTS-strategy.md
@@ -35,4 +35,4 @@
 * Several semver major branches will be maintained at the same time while we have overlapping supported versions
 * No features are backported to older semver major branches
 * Critical bugfixes will be backported to older supported semver major branches
-* High-impact security fixes are very likely to be backported to older semver major branches
+* Security fixes will be backported to older supported semver major branches
--- a/docs/LTS-strategy.md
+++ b/docs/LTS-strategy.md
@@ -17,7 +17,7 @@
 
 * Avoid blocking Node.js core from making progress and removing deprecated parts of the codebase
 * Avoid getting outdated bug reports or security issues
-* Keep options for adopting newer patterns and tooling open
+* Keep options open for adopting newer patterns and tooling
 
 ### Rationale behind conservative policy of breaking changes
 
--- a/docs/LTS-strategy.md
+++ b/docs/LTS-strategy.md
@@ -34,5 +34,5 @@
 
 * Several semver major branches will be maintained at the same time while we have overlapping supported versions
 * No features are backported to older semver major branches
-* Critical bugfixes may be backported to older semver major branches
+* Critical bugfixes will be backported to older supported semver major branches
 * High-impact security fixes are very likely to be backported to older semver major branches
--- a/docs/LTS-strategy.md
+++ b/docs/LTS-strategy.md
@@ -27,7 +27,7 @@
 ### Guidelines for dropping support of Node.js versions
 
 * In order to smoothen the experience of users updating to a new semver major Express version, Express should always support at least one even numbered Node.js version above the current lowest supported one before the drop can be considered. For example, if the lowest supported version is 20, Express support for Node 20 cannot be dropped until Express supports at least Node 22. This ensures that Express updates can be done independently from the Node.js version update.
-* Version support removal must come with clear benefits for one of the three - Express users, Express maintainers or the Node.js core project. Version support is not dropped on a merely time-based basis.
+* Version support removal must come with clear benefits for one of the three - Express users, Express maintainers, or the Node.js core project. Version support is not dropped on a merely time-based basis, or otherwise arbitrarily.
 * Any discontinuation of a Node.js version support is always a semver major release.
 
 ### Active branches
--- a/docs/LTS-strategy.md
+++ b/docs/LTS-strategy.md
@@ -1,0 +1,38 @@
+# Express LTS Strategy
+
+## Goals
+
+* Maintain strict Node.js version support within a single Express major release.
+* Provide good ergonomics and developer experience for Express maintainers
+* Provide good experience for users who want to keep their Express projects that up-to-date
+* Avoid stagnation and blocking innovation that is good for the end users
+
+## Non-Goals
+
+* Only support the active LTS releases of Node.js
+* Support the widest possible set of Node.js versions. We aim to maintain a healthy balance between backwards compatibility and toolchain modernization.
+* Provide timed cadence of Node.js version support dismissal
+
+### Rationale behind dropping support for old Node.js versions
+
+* Avoid blocking Node.js core from making progress and removing deprecated parts of the codebase
+* Avoid getting outdated bug reports or security issues
+* Keep options for adopting newer patterns and tooling open
+
+### Rationale behind conservative policy of breaking changes
+
+* Minimize disruption for end users
+* Be mindful of the fact that some users are still running obsolete versions of Node in production
+
+### Guidelines for dropping support of Node.js versions
+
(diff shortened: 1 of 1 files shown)


docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
+# Express LTS Strategy
+
+## Goals
+
+* Maintain strict Node.js version support within a single Express major release.
+* Provide good ergonomics and developer experience for Express maintainers
+* Provide good experience for users who want to keep their Express projects that up-to-date
+* Avoid stagnation and blocking innovation that is good for the end users
```
wesleytodd, line 8: I think folks might read this as "in order of priority". I know that is not intended, but I wonder if we should move "Provide good experience for users who want to keep their Express projects that up-to-date" to the top?

docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Non-Goals
+
+* Only support the active LTS releases of Node.js
+* Support the widest possible set of Node.js versions. We aim to maintain a healthy balance between backwards compatibility and toolchain modernization.
```
wesleytodd, line 13: Love this section, especially this line! 🎉

docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
...
+* Provide timed cadence of Node.js version support dismissal
+
+### Rationale behind dropping support for old Node.js versions
+
+* Avoid blocking Node.js core from making progress and removing deprecated parts of the codebase
+* Avoid getting outdated bug reports or security issues
+* Keep options for adopting newer patterns and tooling open
```
wesleytodd, line 20:
```suggestion
* Keep options open for adopting newer patterns and tooling
```

docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Rationale behind conservative policy of breaking changes
+
+* Minimize disruption for end users
+* Be mindful of the fact that some users are still running obsolete versions of Node in production
+
+### Guidelines for dropping support of Node.js versions
```
wesleytodd, line 27: I think this is the section I want us to discuss more deeply. Ideally we also put together a calendar view like Node does to see in comparison to Node majors how express majors would line up.

docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
...
+
+* In order to smoothen the experience of users updating to a new semver major Express version, Express should always support at least one even numbered Node.js version above the current lowest supported one before the drop can be considered. For example, if the lowest supported version is 20, Express support for Node 20 cannot be dropped until Express supports at least Node 22. This ensures that Express updates can be done independently from the Node.js version update.
+* Version support removal must come with clear benefits for one of the three - Express users, Express maintainers or the Node.js core project. Version support is not dropped on a merely time-based basis.
+* Any discontinuation of a Node.js version support is always a semver major release.
+
+### Active branches
+
+* Several semver major branches can be maintained at the same time, in case there is a significant userbase for older semver major releases.
```
wesleytodd, line 35:
```suggestion
* Several semver major branches will be maintained at the same time while we have overlapping supported versions
```

docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
...
+* No features are backported to older semver major branches
```
wesleytodd, line 36: Thumbs up on this, but I do think maybe we want to discuss this as a group in more detail.

docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
...
+* Critical bugfixes may be backported to older semver major branches
```
wesleytodd, line 37:
```suggestion
* Critical bugfixes will be backported to older supported semver major branches
```

I think we should commit to this.

docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
...
+* High-impact security fixes are very likely to be backported to older semver major branches
```
wesleytodd, line 38:
```suggestion
* Security fixes will be backported to older supported semver major branches
```

I also think we should commit to doing this for their support lifetime.

wesleytodd (changes requested):
This is AWESOME. Thanks for getting this started. I made some comments to seed conversation, and I think we might want to expand a bit more on some of them, but generally I think this is an awesome start.

docs/LTS-strategy.md
ljharb: I'm not actually sure this makes sense - a healthy balance imo is the same thing as "widest possible set". If a node version prevents using modern toolchains, for example, then that's not a version that'd be included in that set.

docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
...
+* Keep options for adopting newer patterns and tooling open
+
+### Rationale behind conservative policy of breaking changes
+
+* Minimize disruption for end users
+* Be mindful of the fact that some users are still running obsolete versions of Node in production
+
+### Guidelines for dropping support of Node.js versions
+
+* In order to smoothen the experience of users updating to a new semver major Express version, Express should always support at least one even numbered Node.js version above the current lowest supported one before the drop can be considered. For example, if the lowest supported version is 20, Express support for Node 20 cannot be dropped until Express supports at least Node 22. This ensures that Express updates can be done independently from the Node.js version update.
```
ljharb, line 29: this is great, thanks for capturing it

docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
...
+* Version support removal must come with clear benefits for one of the three - Express users, Express maintainers or the Node.js core project. Version support is not dropped on a merely time-based basis.
```
ljharb, line 30:
```suggestion
* Version support removal must come with clear benefits for one of the three - Express users, Express maintainers, or the Node.js core project. Version support is not dropped on a merely time-based basis, or otherwise arbitrarily.
```

docs/LTS-strategy.md
ljharb: this should probably be done as a matter of course anyways, so that if a security issue pops up that necessitates a backport, all the necessary infra is already in place.

docs/LTS-strategy.md
ljharb: i agree, but i think there should also be "may be backported to unsupported semver major branches" as well.

ljharb (commented):
Thanks for writing this up!

docs/LTS-strategy.md
kibertoad: @ljharb I'll think about possible wording improvements here, but the intent is to emphasize that we only want to support a set of versions that do not contradict the rest of the goals defined. This is very important to make explicit in the context of Express specifically, as it has a long-standing tradition of preserving the compatibility to the detriment of everything else.

docs/LTS-strategy.md
UlisesGascon: Yeah! Having a calendar is one of the best ways to explain the rules for the end users, but currently is very solid 🙂

docs/LTS-strategy.md
UlisesGascon: +1 this is a great way to explain it

docs/LTS-strategy.md
UlisesGascon: Yep, we should backport only fixes, perf improvements and so on.. but agree that new features are out of scope 💯

UlisesGascon (approved):
Fantastic proposal @kibertoad! Thanks for doing this. 

IMO this is great and I feel confident with the current state (previous discussed comments aside). I assume that might do more tweaks in the future (but not a blocker for this PR), specially when we will include the release schedule (calendar) as we will need to fit this ideas more intro a concrete schematics 👍

docs/LTS-strategy.md
ljharb: I don’t think any of the detriments you refer to have anything to do with refusing to drop compat for node, but with other things - but i look forward to an alternative phrasing.

wesleytodd: > a healthy balance imo is the same thing as "widest possible set"

As other said, I disagree with the wording but maybe your meaning hinges on the definition we use for "possible"?

Like if we said "we want to ship diagnostic channels so it is not possible to support versions before them" (I am trying to make a hypnotically based on a reality from the past https://github.com/pillarjs/router/pull/96) does that fit your definition? We would drop the old node versions which dont support it in the next major so we could ship support without a lot of compatibility code.

docs/LTS-strategy.md
wesleytodd: I think we should make a diagram like node has, maybe even overlay our support *over* the existing node support schedules?

docs/LTS-strategy.md
ljharb: Yes, of course it fits my definition. If we need a feature that requires dropping a version, then that version is no longer possible to support.

I also especially like the implied path of "ship support with compat code, and then drop the compat code and the node versions that need it in the next major" - that's a *very* user-friendly way to do a major bump.

wesleytodd:
Here is an example we can also reference. https://github.com/debug-js/debug/releases?page=2

Debug, a direct dep, dropped node versions which blocked us updating.

docs/LTS-strategy.md
kibertoad: addressed

docs/LTS-strategy.md
kibertoad: @wesleytodd @UlisesGascon I think @ljharb strongly objected to having a timed deprecation without a practical justification. We need to have this established unambiguously within the strategy. Do we or not do time schedule-based Node version deprecation?

docs/LTS-strategy.md
kibertoad: let's revise in a separate PR if needed

docs/LTS-strategy.md
```diff
@@ -0,0 +1,38 @@
...
+* Maintain strict Node.js version support within a single Express major release
+* Provide good experience for users who want to keep their Express projects that up-to-date
+* Avoid stagnation and blocking innovation that is good for the end users
+* Ensure good ergonomics and developer experience for Express maintainers
+
+## Non-Goals
+
+* Only support the active LTS releases of Node.js
+* Provide timed cadence of Node.js version support dismissal
+* Support an extended set of Node.js versions to the detriment of all other factors. We aim to maintain a healthy balance between backwards compatibility and toolchain modernization
```
kibertoad, line 14: @ljharb @wesleytodd @UlisesGascon Please review the updated wording

ljharb: How about this:
```suggestion
* Prioritizing "support older Node.js versions" over other factors. We aim to maintain a healthy balance between backwards compatibility and toolchain modernization
```

kibertoad: sure, that works!

ljharb (commented):
LGTM! thanks for iterating

kibertoad:
OK, I think this might be ready to be merged as a first version, but of course we should keep iterating over it.

Should we open the discussion of what toolchain to use as a next step? This would help inform the decision of which particular versions to support in the first upcoming semver major release

docs/LTS-strategy.md
wesleytodd: Can we merge without a consensus on this? I am leaning yes? And then we can open targeted PRs to update these and discuss after?

wesleytodd (approved):

docs/LTS-strategy.md
ljharb: If it’s worded as something like “if a release is warranted, this is the schedule when it will happen - but if not, no release will happen” then my concern is gone.

ljharb: Either way, but it seems easy enough to do it before merging ¯\\\_(ツ)_/¯

wesleytodd: @kibertoad if you get to it, awesome. If not, lets try and merge this before EOD so we can start those other PRs to clarify and expand.

kibertoad: addressed. unless there are major outstanding objections to some of the wording, I would suggest merging it as-is and iterate in follow-up PRs.

@wesleytodd Can I get push permissions for the repo?

kibertoad: thanks!

PR #206, opened 2024-03-02
feat: use pkgjs/meet to generate meeting agenda

wesleytodd:
I added support in `pkgjs/meet` for generating meeting agendas. I hope this works well, I tested in a personal repo and it seems to. We will want to edit the template and stuff with the right links but I wanted to make sure this was all working first.

### DIFF
--- a/.github/ISSUE_TEMPLATE/meeting-tc.md
+++ b/.github/ISSUE_TEMPLATE/meeting-tc.md
@@ -1,0 +1,43 @@
+## Date/Time
+
+| Timezone | Date/Time |
+|----------|-----------|
+<%= [
+  'America/Los_Angeles',
+  'America/Denver',
+  'America/Chicago',
+  'America/New_York',
+  'Europe/London',
+  'Europe/Amsterdam',
+  'Europe/Moscow',
+  'Asia/Kolkata',
+  'Asia/Shanghai',
+  'Asia/Tokyo',
+  'Australia/Sydney'
+].map((zone) => {
+  return `| ${zone} | ${date.setZone(zone).toFormat('EEE dd-MMM-yyyy HH:mm (hh:mm a)')} |`
+}).join('\n') %>
+
+Or in your local time:
+* https://www.timeanddate.com/worldclock/?iso=<%= date.toFormat("yyyy-MM-dd'T'HH:mm:ss") %>
+
+## Agenda
+
+Extracted from **<%= agendaLabel %>** labelled issues and pull requests from **<%= owner %>/<%= repo %>** prior to the meeting.
+
+<%= agendaIssues.map((i) => {
+  return `* ${i.html_url}`
+}).join('\n') %>
+
+## Invited
+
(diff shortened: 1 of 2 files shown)


wesleytodd:
Ok, fixed the template but now it is creating a new issue even though it is supposed to be updating the existing one. Next fix in the works. Sorry for the noise.

.github/ISSUE_TEMPLATE/meeting-tc.md
```diff
@@ -0,0 +1,43 @@
...
+Or in your local time:
+* https://www.timeanddate.com/worldclock/?iso=<%= date.toFormat("yyyy-MM-dd'T'HH:mm:ss") %>
+
+## Agenda
+
+Extracted from **<%= agendaLabel %>** labelled issues and pull requests from **<%= owner %>/<%= repo %>** prior to the meeting.
+
+<%= agendaIssues.map((i) => {
+  return `* ${i.html_url}`
+}).join('\n') %>
```
UlisesGascon, line 30: I think that we might want to prevent duplicate items, for reference (https://github.com/expressjs/discussions/issues/208):

Screenshot 2024-03-04 at 01 12 05

.github/workflows/meetings.yml
```diff
@@ -0,0 +1,24 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Technical Committee
+        uses: 'pkgjs/meet@v0'
+        with:
+          issueTitle: '<%= date.toFormat("yyyy-MM-dd") %> Express TC Meeting'
+          token: ${{ secrets.GITHUB_TOKEN }}
+          orgs: expressjs,pillarjs,jshttp
+          agendaLabel: 'top priority'
```
jonchurch, line 19: I'd expect this to be `agenda` or an array that allows `['top-priority', 'agenda']`

Adding an `agenda` label is the convention I'd expect. Easy to slap an agenda label on something to know it will show up in the meeting. Not everything to be discussed is always going to be "top priority"

jonchurch (approved):

.github/ISSUE_TEMPLATE/meeting-tc.md
wesleytodd: I would rather do it in the action than the template. Should be easy enough to add. I'm not sure how the code in place ended with duplicates but it's not a big deal to ignore that unexpected behavior with an explicit check to remove the dupes.

.github/workflows/meetings.yml
wesleytodd: Yeah I only used this to start because it was preexisting. We can go through all of them and add a better label for future.

.github/ISSUE_TEMPLATE/meeting-tc.md
wesleytodd: I think I fixed that in https://github.com/pkgjs/meet/pull/202

PR #283, opened 2024-10-15
Update working session link

bjohansebas:


bjohansebas:
Although maybe there is an option in `pkgs/meet` so that the template doesn't have to be duplicated?

jonchurch:
Ah good to know.

Maintaining two nearly identical templates for a single link difference isn't ideal here.

We can update the pkgjs action to accept some data via options to pass in. That's a new feature to add.

.github/ISSUE_TEMPLATE/meeting-work-session.md
```diff
@@ -0,0 +1,47 @@
...
+
+This meeting is open for anyone who wants to attend. Reminder to follow our [Code of Conduct](https://github.com/expressjs/express/blob/master/Code-Of-Conduct.md).
+
+@expressjs/express-tc
+@expressjs/triagers
+@expressjs/security-wg
+
+## Links
+
+* Minutes:
```
kjugi, line 42: No minutes link?

.github/ISSUE_TEMPLATE/meeting-work-session.md
```diff
@@ -0,0 +1,47 @@
...
+  'America/New_York',
+  'Europe/London',
+  'Europe/Amsterdam',
+  'Europe/Moscow',
+  'Asia/Kolkata',
+  'Asia/Shanghai',
+  'Asia/Tokyo',
+  'Australia/Sydney'
+].map((zone) => {
+  return `| ${zone} | ${date.setZone(zone).toFormat('EEE dd-MMM-yyyy HH:mm (hh:mm a)')} |`
```
kjugi, line 18: Last issue (https://github.com/expressjs/discussions/issues/308) got different hour according to openjs calendar. Should we adjust something there?

.github/ISSUE_TEMPLATE/meeting-work-session.md
bjohansebas: This should be edited by a TC member when the issue is created and added.

jonchurch:
A hack for now could be to check the issue title in the agenda template to determine if it's TC or Working session, and choose the link to use based on that.

jonchurch (changes requested):
Another workaround, for our specific case, would be to check the title in the template. Im pretty sure it gets passed into the template as a `title`

We can check that to differentiate between the two types of meetings inside the ejs template and pick a meeting link that way.

.github/ISSUE_TEMPLATE/meeting-work-session.md
bjohansebas: fixed

kjugi (approved):

wesleytodd (approved):
I didnt check the time changes, but I assume these are reflective of the correct times or they wouldn't have been updated? Should we update the comment above? Is this because it is not taking into account DST?

bjohansebas:
Yes, the time change is to reflect the correct time, which is now one hour later, just like it was before #253.

bjohansebas:
I've created a PR to add the option and avoid implementing the proposed hack

see: https://github.com/pkgjs/meet/pull/207

UlisesGascon (approved):

bjohansebas:
Once https://github.com/pkgjs/meet/pull/207 is released, this can be merged

wesleytodd:
Just merged that now. @jonchurch has a change request on this, not sure if it is still valid?

bjohansebas:
I think we can merge this, since that "request changes" was because I was duplicating the meeting template, which led to the current solution

wesleytodd:
@jonchurch if that is the case, can you swap to an approve? We are not in a rush so can wait for him to reply.

jonchurch (approved):

PR #284, opened 2024-10-15
add triage and security meetings

bjohansebas:


bjohansebas:
I am going to close this issue because I can't find a way to make the workflow work properly, it always gives an error, and it doesn't give me the option to convert this PR into a draft. Once the schedules are defined in the first meetings, if no one else has added the meetings, I will reopen this PR and try again until it works.

UlisesGascon (approved):

.github/workflows/meetings.yml
```diff
@@ -38,3 +38,16 @@ jobs:
...
+        with:
+            issueTitle: '<%= date.toFormat("yyyy-MM-dd") %> Express Triage Session'
+            token: ${{ secrets.GITHUB_TOKEN }}
+            orgs: expressjs,pillarjs,jshttp
+            agendaLabel: 'triage-meeting'
+            meetingLabels: 'meeting'
+            # https://github.com/expressjs/discussions/issues/276#issuecomment-2399741373
+            # Starting on 2024-10-28 at 7pm UTC (2024-10-28T19:00:00.0Z) with a period of 4 weeks (P4W)
+            schedules: '2024-10-28T19:00:00.0Z/P4W'
+            createWithin: 'P1W'
```
UlisesGascon, line 52: Not sure if this is correct :thinking:. So far seems like once a month in this slot is the current agreement.

bjohansebas: Yeah, but this will create the issue a week before, I don’t think we want to create it a month in advance.

UlisesGascon: ahhh! I see, I was thinking that his is the frequency of the meeting :face_in_clouds:

rxmarbles (commented):
LGTM

wesleytodd (approved):
I didn't double check the calendar, but assuming this is correct LGTM.

bjohansebas:
I think we could unify the meeting template once #283 is launched.

rxmarbles (approved):

UlisesGascon:
are we ok to merge it? :)

bjohansebas:
@UlisesGascon I would prefer to wait for #283 to be released, so we don't have to create a new template and maybe use this new template for the other two meetings.

bjohansebas:
Okay, the meeting actually happens on the 3rd Monday of the month, so this automation wouldn't work. I'll have to figure out how to create it for every 3rd Monday of the month, or @wes or @ulies would need to modify the meeting recurrence since they are the only ones I know who have access to change it. The same applies to the security meeting.

image

bjohansebas:
We can't merge this until the recurrence of the meeting is fixed, as it's currently set to the 4th Monday of the month, but it should be every 4 weeks. Additionally, `Meet` doesn't support the option to create the meeting on the 4th or 3rd Monday. @wesleytodd needs to fix the recurrence.

wesleytodd:
What specifically do I need to fix? Is this in the LFX platform or on the calendar? Can you help clarify?

bjohansebas:
It's on the LFX platform, so that the recurrence is every 4 weeks instead of the last Monday of each month.

wesleytodd:
Screenshot 2025-02-10 at 4 13 28 PM

I just updated it to this. Is this correct?

bjohansebas:
I think it can be merged now, I changed the date to make sure it gets created every 4 week, I hope it's correct.

The error is because it's trying to create a new issue but doesn't have the permissions to do so since it's in a PR.

Issue #285, opened 2024-10-16
Re-thinking the discussion process


UlisesGascon:
I love this space that we have to discuss and bring new ideas to the table. So far, it's working well, but I believe we can make some small changes to make it even **more awesome<sup>TM</sup>**.

**Current Challenges**

So far, once a discussion is solved and we arrive at a conclusion, we don't always provide a summary. This can result in re-entering the same discussion later, or we forget the last agreement made, causing us to revisit the same issue. This is especially problematic if different people are involved in the repeated discussions, as they may miss important context from the previous conversation.

We also have dozens of issues in this repo that are in an unknown state, so we might want to use proper labels to represent different stages of the discussions. Of course, we should also do some cleanup and close a few of them.

**Potential Improvements**

For decision-making, we could use a light implementation of the [Architecture Decision Record (ADR)](https://github.com/joelparkerhenderson/architecture-decision-record). We can store these decisions as markdown files in this repo and revisit them when needed. The goal is to record the motivations behind decisions, the alternatives we explored, etc., so new contributors can have a better understanding of the project's history. I can create a basic example if we agree this is a good idea.

For labeling, I propose the following:

- `New` – For newly created issues
- `Waiting` – When discussions are blocked due to external factors or waiting for POCs, etc.
- `Active` – For discussions currently in progress
- `Waiting for ADR` – Waiting for an ADR PR
- `Discarded` – The proposal did not succeed

cc: @expressjs/triagers @expressjs/express-tc @expressjs/security-wg

PaulaPaul:
In my consulting jobs I always introduce ADRs! 
It's good to have a process for a regular review of completed ADRs (every 3-6 months, or longer, depending on how many decision records are created). That is a good way to learn from the decisions made months/years ago. It can be helpful to revisit those decisions to record 'lessons learned'. Sometimes that results in new decision records that refer back to the original decision.

This process is very valuable for onboarding, and to remind maintainers what we were thinking months (or years) ago. 

Some tips:
- Always define the 'scope' of the ADRs (for example 'These Repos...' or 'all activities that support the xxxxx Project', or 'everything within this Org, except for xxxx')
- Define 'who can create an ADR'. It can be helpful to state that anyone can propose an ADR, and define the process by which it becomes final / merged
- Define some kind of schedule for regular discussion of in-process ADRs (depending on how many you generate, and how urgent), or define the way that people can request a live discussion
[...]

PR #289, opened 2024-10-22
Add policy to align engines fields with ci

### LINKED ISSUE #286: Discussion: Using engines in the package.json
Currently some packages has drop support to Node.js prior to v18, seems like we didn't upgrade the engines field in the `package.json` discordantly.

So, I will like to discuss if this a policy that we should follow as an organization to all the packages (so in that case we can also document it).

:warning: So far seems like Express actually updated the package ([ref](https://github.com/expressjs/express/blob/master/package.json#L83)) but not in others like finalhandler ([ref](https://github.com/pillarjs/finalhandler/blob/master/package.json#L38))

Ref:
- https://github.com/pillarjs/finalhandler/issues/64
- https://github.com/pillarjs/finalhandler/pull/59 (great context on why not update by @wesleytodd )

cc: @expressjs/express-tc @Phillip9587

UlisesGascon:
- closes: https://github.com/expressjs/discussions/issues/286
- related: https://github.com/expressjs/discussions/issues/285

### DIFF
--- a/docs/adr/001-aligning-engines-field-with-ci.md
+++ b/docs/adr/001-aligning-engines-field-with-ci.md
@@ -27,6 +27,7 @@ We will establish a policy that requires the `engines` field in `package.json` t
 
 **What will be done?**  
 - Update the `engines` field in `package.json` for each package to reflect the minimum Node.js version supported by the corresponding CI configuration.
+- Include the `engines` fields in all the projects, so it will clarify what is supported for the final user
 
 **What will not be done?**  
 - We will not support versions of Node.js that are not explicitly defined in the `engines` field or tested in CI.
--- a/docs/adr/001-aligning-engines-field-with-ci.md
+++ b/docs/adr/001-aligning-engines-field-with-ci.md
@@ -1,0 +1,69 @@
+# ADR 001: Aligning `engines` Field in `package.json` with CI Version Support
+
+## Status
+Proposed
+
+## Submitters
+- Ulises Gascón (@UlisesGascon)
+
+## Decision Owners
+- Express TC (@expressjs/express-tc)
+
+## Context
+The issue of maintaining consistency between the `engines` field in `package.json` and the versions tested in CI has been raised. Several packages have dropped support for Node.js versions prior to v18, but this change has not been consistently applied across all packages.
+
+**Why do we need this decision?**  
+Aligning the `engines` field with CI ensures that we explicitly define the minimum Node.js version supported for each release. This minimizes the risk of unintentional support for older versions that are not tested in CI, thus reducing maintenance overhead and unexpected bug reports.
+
+**What problem does it solve or avoid?**  
+This decision will help avoid introducing breaking changes unintentionally, streamline user expectations, and ensure that end users are aware of the minimum Node.js version they should use.
+
+**Are there any existing issues/discussions/pull requests related to this?**  
+- [Discussion: Using engines in the package.json #286](https://github.com/expressjs/discussions/issues/286)
+- [Missing engines Field Update for Node.js Version Support in v2.0.0 pillarjs/finalhandler#64](https://github.com/pillarjs/finalhandler/issues/64)
+
+## Decision
+We will establish a policy that requires the `engines` field in `package.json` to match the Node.js versions defined in our CI configuration. This policy will apply to all packages within the Express ecosystem.
+
+**What will be done?**  
+- Update the `engines` field in `package.json` for each package to reflect the minimum Node.js version supported by the corresponding CI configuration.
+
+**What will not be done?**  
+- We will not support versions of Node.js that are not explicitly defined in the `engines` field or tested in CI.
+
+## Rationale
+
+**Alternatives Considered:**
+- **Alternative 1:** Keep the `engines` field as-is and update only on a need basis.  
+  - **Reason for rejection:** This could lead to confusion among users and contributors regarding the supported Node.js versions, resulting in unintended support for older versions.
+- **Alternative 2:** Remove the `engines` field altogether and rely solely on documentation for version support.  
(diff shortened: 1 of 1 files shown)


bjohansebas (approved):

blakeembrey (approved):

docs/adr/001-aligning-engines-field-with-ci.md
```diff
@@ -0,0 +1,69 @@
...
+**Are there any existing issues/discussions/pull requests related to this?**  
+- [Discussion: Using engines in the package.json #286](https://github.com/expressjs/discussions/issues/286)
+- [Missing engines Field Update for Node.js Version Support in v2.0.0 pillarjs/finalhandler#64](https://github.com/pillarjs/finalhandler/issues/64)
+
+## Decision
+We will establish a policy that requires the `engines` field in `package.json` to match the Node.js versions defined in our CI configuration. This policy will apply to all packages within the Express ecosystem.
+
+**What will be done?**  
+- Update the `engines` field in `package.json` for each package to reflect the minimum Node.js version supported by the corresponding CI configuration.
+
```
UlisesGascon, line 30:
```suggestion
- Include the `engines` fields in all the projects, so it will clarify what is supported for the final user

```

suggested by @ctcpip on the last TC Meeting

rxmarbles (commented):
LGTM

ctcpip:
we need a note in here about the possibility of continuing to run CI in older unsupported node versions so we know if/when a change clearly breaks back-compat. (we do this for express itself, for example)

UlisesGascon:
ping @ctcpip

wesleytodd (approved):
While I have general concerns about the decision here, I would much prefer this decision over the current state. The only thing I think is missing is the call-out that all *more restrictive* engines changes can only land with a *MAJOR*. And that we agreed major versions should not be cut *just to change node support*.  

This means that this policy will take a while to roll out. I do not consider this blocking to merge this ADR, but I think it is something we had agreed on in a few discussions so just wanted to call it out since I didn't see it mentioned in the text of the ADR.

blakeembrey (changes requested):
As mentioned in https://github.com/jshttp/content-disposition/pull/79#issuecomment-2851329726, this proposal doesn’t cover what happens when the engines is incorrect or needs to be added for the first time. I don’t think that belongs in this ADR specifically since it’s about versioning itself vs engines, but it would help to clarify how engines impacts versioning. I.e. if we are clarifying an existing engines support, is it a breaking change?

If we’re releasing majors for only engines changes, this doubles maintenance overhead if there happens to be any issues that need to be back ported to the version with and without engines.

I can think of two options here:

1. Consider fixing the engines field a non-major change (it’s a fix after all)
2. Document the process of adding it to the “next major”

IIRC we discussed that adjusting engines itself should not be a reason to do a major release, but I can’t recall where that conversation is now.

ljharb:
If a thing worked before, and stops working, it's a breaking change. Adding `engines` _can_ be a breaking change even if it's exactly matching the previously working range, because there's tools that check for it in a graph and can fail CI.

voxpelli:
I think the pragmatic thing is to see the `"engines"` as a best effort description at describing the intended compatibility of the code at the point of release.

If the description is out of sync with reality then either reality has to adapt to the description or the other way around.

If the change of the `"engines"` range can be seen as a bug fix – reality was clearly different to it – then I think it's correct to fix, else I think a new major or a widened reality is needed.

Bug fixes are always breaking for those reliant on the bugged behavior and it's always a balance of factors where it's primarily a fix or a breaking change.

ljharb:
Removing `engines` entirely would be explicitly declaring that you support `*` (the documented default), so I'm not sure why that'd be desirable, but certainly that's not a breaking change.

blakeembrey:
> I'm not sure why that'd be desirable, but certainly that's not a breaking change.

It’s not, but it’s less desirable to try and release new major versions to fix the supported version range.

@voxpelli I agree, it’s the main point of contention. We need to document this somewhere.

docs/adr/001-aligning-engines-field-with-ci.md
```diff
@@ -0,0 +1,70 @@
...
+**Are there any existing issues/discussions/pull requests related to this?**  
+- [Discussion: Using engines in the package.json #286](https://github.com/expressjs/discussions/issues/286)
+- [Missing engines Field Update for Node.js Version Support in v2.0.0 pillarjs/finalhandler#64](https://github.com/pillarjs/finalhandler/issues/64)
+
+## Decision
+We will establish a policy that requires the `engines` field in `package.json` to match the Node.js versions defined in our CI configuration. This policy will apply to all packages within the Express ecosystem.
+
+**What will be done?**  
+- Update the `engines` field in `package.json` for each package to reflect the minimum Node.js version supported by the corresponding CI configuration.
+- Include the `engines` fields in all the projects, so it will clarify what is supported for the final user
```
bjohansebas, line 30: @blakeembrey It can be handled in a different ADR
```suggestion
- Include the `engines` fields in all the projects, so it will clarify what is supported for the final user.
- For cases where the engines field was not updated by mistake and Node.js support was explicitly changed in the history of each package, it will be handled in a different ADR.
```

jonchurch:
Notes for me summarizing while ulises talks about this bc it has been so long and I have to do this everytime:

We have engines field in our packages. We have a challenge w/ engines, the problem is if we forget to update the engines field (this has happened, I jon have done this), do we do a new major to fix the engines field? That was an open question, which has ended up being discussed in this ADR as well. 

Its turned into a discussion into whether or not adding/fixing engines is always a major.

Im taking an action item to work with Ulises to caught back up on this and get a clear next step.

ctcpip:
relevant: https://github.com/jshttp/mime-types/pull/136

ljharb:
Fixing `engines` is not a major _if the implementation was already incompatible with the engines being excluded_.

If that's the case, it's a patch! If that's not the case, then I'd suggest reconsidering changing `engines` in the first place :-)

Issue #279, opened 2024-10-08
Discussion: Using caret (`^`) with our own dependencies


UlisesGascon:
Historically, we have avoided the use of caret in the `package.json` files for our own projects. I am just wondering if we want to keep this policy or change it. Anyway, I would like to use this issue to document that decision and to have all the discussion around it in one place.

Ref: https://github.com/expressjs/express/pull/6017#discussion_r1789129519

kurt-apple:
Via semver.org:

> Given a version number MAJOR.MINOR.PATCH, increment the:
> 1. MAJOR version when you make incompatible API changes
> 2. MINOR version when you add functionality in a backward compatible manner
> 3. PATCH version when you make backward compatible bug fixes
> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

In my opinion, pinning to a specific patch version is as extreme as using a ">" or "latest" in package.json. In one extreme, breaking/incompatible changes by definition will be blindly accepted. In another extreme, patches that are backward compatible are rejected and you have to ask yourself, why?

There is an edge case if the dependencies of express have an unexpected regression or decide to throw semver out the window. I think it was noted in the linked PR that express has chosen dependencies that strictly follow semver. When this is the case, then accepting patches from deps _is_ strictly abiding by semver specifications. See semver.org for a more detailed summary of the advantages of doing so.

kurt-apple:
Well said, @dpopp07 
I have definitely experienced unexpected regressions or incompatibilities with my dependencies...
Perhaps the best of both worlds is to be notified of updated packages and prioritize response based on the context/nature of the updated code, reliability of test coverage, and bandwidth to manually review dependencies. Something like dependabot comes to mind, but I look forward to learning more about this myself.

ljharb:
Those risks should be mitigated by end users, using lockfiles - not by intervening packages. Every dependency should always use `^` (unless it doesn't follow semver, or needs to be pinned due to bugs a la `=`).

klaude:
Another random developer piping in! There are two takes that support using carets or tildes. 

Philosophically, explicit version pinning is the downstream project's responsibility. Libraries and frameworks should use carets to express the features they need by their dependencies' major and minor versions. Express' users should pin dependencies to satisfy their security and compliance requirements, but express should keep it looser to allow for the most use cases. 

[...]

PR #290, opened 2024-10-22
Add policy on using caret (^) or tilde (~) with dependencies

UlisesGascon:
Policy on using caret (`^`) or tilde (`~`) with dependencies

- Establishes policy for using caret (`^`) or tilde (`~`) for dependencies that we maintain while avoiding its use for third-party dependencies.
- Explains the differences between caret (^) and tilde (~) version ranges.
- Documents the rationale, alternatives considered, and implementation plan.

**Context**
- closes https://github.com/expressjs/discussions/issues/279
- related https://github.com/expressjs/discussions/issues/285

### DIFF
--- a/docs/adr/002-policy-on-using-caret-with-dependencies.md
+++ b/docs/adr/002-policy-on-using-caret-with-dependencies.md
@@ -23,7 +23,7 @@ This decision aims to reduce the maintenance burden of frequently updating pinne
 - [expressjs/express#6017 (comment)](https://github.com/expressjs/express/issues/6017)
 
 ## Decision
-We will adopt a policy where the caret (`^`) symbol is used for dependencies owned within the Express ecosystem (e.g., `body-parser` for Express), but continue to avoid using it for third-party dependencies that we do not maintain.
+We will adopt a policy where the caret (`^`) symbol is used for dependencies all dependencies.
 
 **What will be done?**  
 - Update the `package.json` files to use `^` for all dependencies (both prod and dev).
--- a/docs/adr/002-policy-on-using-caret-with-dependencies.md
+++ b/docs/adr/002-policy-on-using-caret-with-dependencies.md
@@ -40,9 +40,7 @@ We will adopt a policy where the caret (`^`) symbol is used for dependencies all
 ## Rationale
 
 **Alternatives Considered:**
-- **Alternative 1:** Use `^` for all dependencies, including third-party libraries.  
-- **Reason for rejection:** This could increase the risk of unintended regressions and security issues from third-party updates.
-- **Alternative 2:** Continue pinning all dependencies, including internal ones, to specific versions.  
+- **Alternative:** Continue pinning all dependencies, including internal ones, to specific versions.  
 - **Reason for rejection:** This approach requires frequent updates and increases the maintenance burden, as each minor or patch update requires a new release.
 
 **Pros and Cons**:
--- a/docs/adr/002-policy-on-using-caret-with-dependencies.md
+++ b/docs/adr/002-policy-on-using-caret-with-dependencies.md
@@ -36,7 +36,7 @@ We will adopt a policy where the caret (`^`) symbol is used for dependencies own
 
 **What will not be done?**  
 - We will not use `^` or `~` for external dependencies that are not maintained by the Express organization, as they could introduce unexpected changes.
-- We won't force to use `^` or `~` for own dependencies if there is a reason to use a fix version and it is properly documented.
+- We won't force to use `^` or `~` for own dependencies if there is a reason to use a pinned version and it is properly documented.
 
 ## Rationale
 
--- a/docs/adr/002-policy-on-using-caret-with-dependencies.md
+++ b/docs/adr/002-policy-on-using-caret-with-dependencies.md
@@ -27,7 +27,6 @@ We will adopt a policy where the caret (`^`) symbol is used for dependencies own
 
 **What will be done?**  
 - Update the `package.json` files to use `^` for all dependencies (both prod and dev).
-- Continue using fixed versions (no caret or tilde) for external dependencies to minimize risks of unintended updates.
 
 **Note on `^` vs. `~`:**  
 - `^` allows updates to the most recent minor or patch version, offering greater flexibility and reducing the need for frequent manual updates. For example, `^1.2.3` will accept updates to `1.3.0`, `1.4.0`, but not `2.0.0`.
--- a/docs/adr/002-policy-on-using-caret-with-dependencies.md
+++ b/docs/adr/002-policy-on-using-caret-with-dependencies.md
@@ -26,7 +26,7 @@ This decision aims to avoid unexpected updates and regressions from external dep
 We will adopt a policy where the caret (`^`) symbol is used for dependencies owned within the Express ecosystem (e.g., `body-parser` for Express), but continue to avoid using it for third-party dependencies that we do not maintain.
 
 **What will be done?**  
-- Update the `package.json` files to use `^` for our own dependencies.
+- Update the `package.json` files to use `^` for all dependencies (both prod and dev).
 - Continue using fixed versions (no caret or tilde) for external dependencies to minimize risks of unintended updates.
 
 **Note on `^` vs. `~`:**  
--- a/docs/adr/002-policy-on-using-caret-with-dependencies.md
+++ b/docs/adr/002-policy-on-using-caret-with-dependencies.md
@@ -30,8 +30,7 @@ We will adopt a policy where the caret (`^`) symbol is used for dependencies all
 
 **Note on `^` vs. `~`:**  
 - `^` allows updates to the most recent minor or patch version, offering greater flexibility and reducing the need for frequent manual updates. For example, `^1.2.3` will accept updates to `1.3.0`, `1.4.0`, but not `2.0.0`.
-- `~` is more conservative, only allowing updates to patch versions. For example, `~1.2.3` will accept updates to `1.2.4`, `1.2.5`, but not `1.3.0`.
-- For our own dependencies that adhere strictly to semver, `^` is preferred over `~`, while for third-party libraries, a stricter versioning strategy is maintained to prevent unexpected regressions.
+- `~` is more conservative, only allowing updates to patch versions. For example, `~1.2.3` will accept updates to `1.2.4`, `1.2.5`, but not `1.3.0`. Some Node.js legacy versions only support this, like Node@0.8.
 
 **What will not be done?**  
 - We won't force to use `^` or `~` for any dependency if there is a reason to use a pinned version and it is properly documented.
--- a/docs/adr/002-policy-on-using-caret-with-dependencies.md
+++ b/docs/adr/002-policy-on-using-caret-with-dependencies.md
@@ -72,7 +72,7 @@ This decision strikes a balance between reducing maintenance effort and managing
 
 ## Implementation
 
-- **Phase 1**: Update `package.json` files across all packages to apply `^` for all dependencies.
+- **Phase 1**: Update `package.json` files across all packages to apply `^` for all dependencies which support greater than node@0.10.0.
 - **Phase 2**: Review and adjust documentation to include the new policy on dependency versioning.
 
 ## References
--- a/docs/adr/002-policy-on-using-caret-with-dependencies.md
+++ b/docs/adr/002-policy-on-using-caret-with-dependencies.md
@@ -16,7 +16,7 @@ Historically, the Express project has avoided using the caret (`^`) in the `pack
 Clarifying the policy on using caret (`^`) helps to ensure consistency across the Express ecosystem, reduce unnecessary maintenance, and prevent unexpected dependency updates. It also addresses concerns about the balance between keeping dependencies up-to-date and avoiding regressions.
 
 **What problem does it solve or avoid?**  
-This decision aims to avoid unexpected updates and regressions from external dependencies while reducing the maintenance burden of frequently updating pinned dependencies.
+This decision aims to reduce the maintenance burden of frequently updating pinned dependencies.
 
 **Are there any existing issues/discussions/pull requests related to this?**  
 - [Discussion: Using caret (^) with our own dependencies #279](https://github.com/expressjs/discussions/issues/279)
--- a/docs/adr/002-policy-on-using-caret-with-dependencies.md
+++ b/docs/adr/002-policy-on-using-caret-with-dependencies.md
@@ -1,0 +1,87 @@
+# ADR 002: Policy on Using Caret (`^`) or tilde (`~`) with Dependencies in `package.json`
+
+## Status
+Proposed
+
+## Submitters
+- Ulises Gascón (@UlisesGascon)
+
+## Decision Owners
+- Express TC (@expressjs/express-tc)
+
+## Context
+Historically, the Express project has avoided using the caret (`^`) in the `package.json` files for its own dependencies. This decision aims to review whether this practice should continue or if adjustments are needed.
+
+**Why do we need this decision?**  
+Clarifying the policy on using caret (`^`) helps to ensure consistency across the Express ecosystem, reduce unnecessary maintenance, and prevent unexpected dependency updates. It also addresses concerns about the balance between keeping dependencies up-to-date and avoiding regressions.
+
+**What problem does it solve or avoid?**  
+This decision aims to avoid unexpected updates and regressions from external dependencies while reducing the maintenance burden of frequently updating pinned dependencies.
+
+**Are there any existing issues/discussions/pull requests related to this?**  
+- [Discussion: Using caret (^) with our own dependencies #279](https://github.com/expressjs/discussions/issues/279)
+- [expressjs/express#6017 (comment)](https://github.com/expressjs/express/issues/6017)
+
+## Decision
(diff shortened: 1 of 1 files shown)


docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,87 @@
...
+
+**What problem does it solve or avoid?**  
+This decision aims to avoid unexpected updates and regressions from external dependencies while reducing the maintenance burden of frequently updating pinned dependencies.
+
+**Are there any existing issues/discussions/pull requests related to this?**  
+- [Discussion: Using caret (^) with our own dependencies #279](https://github.com/expressjs/discussions/issues/279)
+- [expressjs/express#6017 (comment)](https://github.com/expressjs/express/issues/6017)
+
+## Decision
+We will adopt a policy where the caret (`^`) symbol is used for dependencies owned within the Express ecosystem (e.g., `body-parser` for Express), but continue to avoid using it for third-party dependencies that we do not maintain.
```
ljharb, line 26: why not also for third party deps? most maintainers don't backport fixes to previous minors.

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,87 @@
...
+
+**What will not be done?**  
+- We will not use `^` or `~` for external dependencies that are not maintained by the Express organization, as they could introduce unexpected changes.
+- We won't force to use `^` or `~` for own dependencies if there is a reason to use a fix version and it is properly documented.
+
+## Rationale
+
+**Alternatives Considered:**
+- **Alternative 1:** Use `^` for all dependencies, including third-party libraries.  
+- **Reason for rejection:** This could increase the risk of unintended regressions and security issues from third-party updates.
```
ljharb, line 45: imo this risk is lower than the risk of dependency vulnerability/regression *fixes* from *not* being available to downstream users.

docs/adr/002-policy-on-using-caret-with-dependencies.md
blakeembrey: Agreed, we should always use `^` and use dependencies/authors we trust, no need to fight the ecosystem.

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,87 @@
...
+- Update the `package.json` files to use `^` for our own dependencies.
+- Continue using fixed versions (no caret or tilde) for external dependencies to minimize risks of unintended updates.
+
+**Note on `^` vs. `~`:**  
+- `^` allows updates to the most recent minor or patch version, offering greater flexibility and reducing the need for frequent manual updates. For example, `^1.2.3` will accept updates to `1.3.0`, `1.4.0`, but not `2.0.0`.
+- `~` is more conservative, only allowing updates to patch versions. For example, `~1.2.3` will accept updates to `1.2.4`, `1.2.5`, but not `1.3.0`.
+- For our own dependencies that adhere strictly to semver, `^` is preferred over `~`, while for third-party libraries, a stricter versioning strategy is maintained to prevent unexpected regressions.
+
+**What will not be done?**  
+- We will not use `^` or `~` for external dependencies that are not maintained by the Express organization, as they could introduce unexpected changes.
```
blakeembrey, line 38: I'm not sure there's any reason to use `~` either, I think we would have used it instead of `^` due to supported NPM versions.

blakeembrey (commented):
TL;DR I think we should try to be high trust and good examples of OSS by treating all dependencies the same.

docs/adr/002-policy-on-using-caret-with-dependencies.md
ljharb: That's indeed the only reason express used `~`, because v4 came out at a time when not every npm version supported `^`.

wesleytodd: Yep, I think we can remove this whole line.

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,87 @@
...
+Proposed
+
+## Submitters
+- Ulises Gascón (@UlisesGascon)
+
+## Decision Owners
+- Express TC (@expressjs/express-tc)
+
+## Context
+Historically, the Express project has avoided using the caret (`^`) in the `package.json` files for its own dependencies. This decision aims to review whether this practice should continue or if adjustments are needed.
```
wesleytodd, line 13:
```suggestion
Historically, the Express project has avoided using the caret (`^`) in the `package.json` files for its own dependencies. This proposal aims to review whether this practice should continue or if adjustments are needed.
```

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,87 @@
...
+
+**Why do we need this decision?**  
+Clarifying the policy on using caret (`^`) helps to ensure consistency across the Express ecosystem, reduce unnecessary maintenance, and prevent unexpected dependency updates. It also addresses concerns about the balance between keeping dependencies up-to-date and avoiding regressions.
+
+**What problem does it solve or avoid?**  
+This decision aims to avoid unexpected updates and regressions from external dependencies while reducing the maintenance burden of frequently updating pinned dependencies.
```
wesleytodd, line 19:
```suggestion
This decision aims to reduce the maintenance burden of frequently updating pinned dependencies.
```

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,87 @@
...
+
+**Are there any existing issues/discussions/pull requests related to this?**  
+- [Discussion: Using caret (^) with our own dependencies #279](https://github.com/expressjs/discussions/issues/279)
+- [expressjs/express#6017 (comment)](https://github.com/expressjs/express/issues/6017)
+
+## Decision
+We will adopt a policy where the caret (`^`) symbol is used for dependencies owned within the Express ecosystem (e.g., `body-parser` for Express), but continue to avoid using it for third-party dependencies that we do not maintain.
+
+**What will be done?**  
+- Update the `package.json` files to use `^` for our own dependencies.
```
wesleytodd, line 29:
```suggestion
- Update the `package.json` files to use `^` for all dependencies (both prod and dev).
```

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,87 @@
...
+- Continue using fixed versions (no caret or tilde) for external dependencies to minimize risks of unintended updates.
```
wesleytodd, line 30:
```suggestion

```

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,87 @@
...
+- Users relying on strict version control may need to adjust their expectations when using our own dependencies.
+
+**Mitigations**:
+- Strong test coverage and CI checks will help catch potential issues early.
+- Clear communication in documentation and release notes to inform users of the updated dependency policy.
+- Ensure that we are following strict semver when releasing our own libraries.
+
+## Implementation
+
+- **Phase 1**: Update `package.json` files across all packages to apply `^` for our own dependencies.
```
wesleytodd, line 80:
```suggestion
- **Phase 1**: Update `package.json` files across all packages to apply `^` for all dependencies.
```

wesleytodd (changes requested):
If @UlisesGascon is alright with the suggestions to use `^` for all, I tried to make change suggestions to everywhere I noticed where it was referenced. Personally I think the conservative approach used in earlier days is not necessary anymore and so am in favor of using it for all deps.

docs/adr/002-policy-on-using-caret-with-dependencies.md
ljharb:
```suggestion
We will adopt a policy where the caret (`^`) symbol is used for all dependencies.
```

UlisesGascon:
Worth mention that caret is not supported on Node@0.8, see: https://github.com/expressjs/morgan/pull/298#issuecomment-2439662289. So this change might be semver major in many scenarios, unless we can consider dropping support for Node@0.8 as not a major :smile:

ljharb:
@UlisesGascon it's not supported on the npm that comes with node 0.8, but if you use `nvm install-latest-npm` to get the latest one, then it works just fine :-)

either way, i think it can be fine if you need a newer node to _install_ it as long as it works in old node (albeit, not ideal)

bjohansebas:
So we need to prepare the CI for the packages that still support node.js@0.8 before making any changes to the dependencies

ljharb:
Yes, for any package that isn't already declaring engines, the first step should be testing (manual or automated) to determine on which versions it's already working - because those are the engines that it's non-breaking to declare.

bjohansebas:
@expressjs/express-tc, the way forward is to update the CI, or do we prefer to do this in a major version when support for Node.js 0.8 is removed?

wesleytodd:
I think the suggestions are mostly agreed upon, @UlisesGascon if you want to make these I can change my review to approved and we can merge this right?

The rest of the questions/comments dont seem applicable now right?

docs/adr/002-policy-on-using-caret-with-dependencies.md
bjohansebas: it’s going to be all the dependencies, right?

ljharb: As long as the dep follows semver, yes - otherwise it might use `~` or `=` - but those would be one-off cases.

wesleytodd:
> he way forward is to update the CI, or do we prefer to do this in a major version when support for Node.js 0.8 is removed?

Ref: https://github.com/expressjs/compression/pull/214#issuecomment-2596868416

I don't think we should make the CI more complicated by applying this change to packages we have not rev'd to drop node versions. I think it is better to do nothing to those until we have another reason to major rev them.

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,87 @@
...
+- Continue using fixed versions (no caret or tilde) for external dependencies to minimize risks of unintended updates.
+
+**Note on `^` vs. `~`:**  
+- `^` allows updates to the most recent minor or patch version, offering greater flexibility and reducing the need for frequent manual updates. For example, `^1.2.3` will accept updates to `1.3.0`, `1.4.0`, but not `2.0.0`.
+- `~` is more conservative, only allowing updates to patch versions. For example, `~1.2.3` will accept updates to `1.2.4`, `1.2.5`, but not `1.3.0`.
+- For our own dependencies that adhere strictly to semver, `^` is preferred over `~`, while for third-party libraries, a stricter versioning strategy is maintained to prevent unexpected regressions.
+
+**What will not be done?**  
+- We will not use `^` or `~` for external dependencies that are not maintained by the Express organization, as they could introduce unexpected changes.
+- We won't force to use `^` or `~` for own dependencies if there is a reason to use a fix version and it is properly documented.
```
ctcpip, line 39:
```suggestion
- We won't force to use `^` or `~` for own dependencies if there is a reason to use a pinned version and it is properly documented.
```

jonchurch:
Im in favor of caret being the default for all deps, external or not. AKA I want to accept the suggestions to change the proposal to state that.

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,86 @@
...
+## Decision
+We will adopt a policy where the caret (`^`) symbol is used for dependencies all dependencies.
+
+**What will be done?**  
+- Update the `package.json` files to use `^` for all dependencies (both prod and dev).
+
+**Note on `^` vs. `~`:**  
+- `^` allows updates to the most recent minor or patch version, offering greater flexibility and reducing the need for frequent manual updates. For example, `^1.2.3` will accept updates to `1.3.0`, `1.4.0`, but not `2.0.0`.
+- `~` is more conservative, only allowing updates to patch versions. For example, `~1.2.3` will accept updates to `1.2.4`, `1.2.5`, but not `1.3.0`.
+- For our own dependencies that adhere strictly to semver, `^` is preferred over `~`, while for third-party libraries, a stricter versioning strategy is maintained to prevent unexpected regressions.
```
UlisesGascon, line 34:
```suggestion
**Note on `^` vs. `~`:**  
- `^` allows updates to the most recent minor or patch version, offering greater flexibility and reducing the need for frequent manual updates. For example, `^1.2.3` will accept updates to `1.3.0`, `1.4.0`, but not `2.0.0`.
- `~` is more conservative, only allowing updates to patch versions. For example, `~1.2.3` will accept updates to `1.2.4`, `1.2.5`, but not `1.3.0`. Some Node.js legacy versions only support this, like Node@0.8.
```

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,86 @@
...
+
+**What will not be done?**  
+- We will not use `^` or `~` for external dependencies that are not maintained by the Express organization, as they could introduce unexpected changes.
+- We won't force to use `^` or `~` for own dependencies if there is a reason to use a pinned version and it is properly documented.
```
UlisesGascon, line 38:
```suggestion
**What will not be done?**  
- We won't force to use `^` or `~` for any dependency if there is a reason to use a pinned version and it is properly documented.
```

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Rationale
+
+**Alternatives Considered:**
+- **Alternative 1:** Use `^` for all dependencies, including third-party libraries.  
+- **Reason for rejection:** This could increase the risk of unintended regressions and security issues from third-party updates.
+- **Alternative 2:** Continue pinning all dependencies, including internal ones, to specific versions.  
+- **Reason for rejection:** This approach requires frequent updates and increases the maintenance burden, as each minor or patch update requires a new release.
```
UlisesGascon, line 46:
```suggestion
- **Alternative:** Continue pinning all dependencies, including internal ones, to specific versions.  
- **Reason for rejection:** This approach requires frequent updates and increases the maintenance burden, as each minor or patch update requires a new release.
```

UlisesGascon:
ok... I think that this is ready for a final review, please feel free to suggest changes and create new discussions if something is not described as agreed in previous calls.

image

Phillip9587 (approved):

docs/adr/002-policy-on-using-caret-with-dependencies.md
```diff
@@ -0,0 +1,82 @@
...
+- Users relying on strict version control may need to adjust their expectations when using our own dependencies.
+
+**Mitigations**:
+- Strong test coverage and CI checks will help catch potential issues early.
+- Clear communication in documentation and release notes to inform users of the updated dependency policy.
+- Ensure that we are following strict semver when releasing our own libraries.
+
+## Implementation
+
+- **Phase 1**: Update `package.json` files across all packages to apply `^` for all dependencies.
```
bjohansebas, line 75: Only in those packages where Node.js support allows it. *compression* cannot at the moment, and this is also the case for several packages that still support Node.js 0.8.

UlisesGascon: true... I include a reference in https://github.com/expressjs/discussions/blob/adr/dependecies/docs/adr/002-policy-on-using-caret-with-dependencies.md?plain=1#L33, but we can also include one explicit here. WDYT?

bjohansebas: Yep, something more explicit here would be great, since not all packages are in the process of leaving the old Node yet.

ljharb: I’m not sure why it can’t? node 0.8 can still upgrade to an npm that supports carets

UlisesGascon: fair point @ljharb, but that will require for the end user to use a never npm version (not the default one) so it adds an additional step for them in order to use the library. But we can easily assume that users that keep running Node@0.8 in 2025 are use to workarounds :smile:

bjohansebas: Yep, although it would complicate our CI a bit, which I already tried to do for compression, but Wes opposed it.

ljharb: nvm install-latest-npm works perfectly :-p

bjohansebas: I tried it, and it actually didn’t work. I don’t remember why, but I can look into it later.

bjohansebas: Now I remember, it was the `engines` field that npm uses, which doesn’t allow it to be used in an older version of Node.

ljharb: certainly `npm install -g npm` will fail for that and many other reasons; that's why nvm has its own custom logic for installing the latest npm _that works_.

wesleytodd:
```suggestion
- **Phase 1**: Update `package.json` files across all packages to apply `^` for all dependencies which support greater than node@0.10.0.
```

wesleytodd (approved):

PR #291, opened 2024-10-22
docs: Add ADR template

UlisesGascon:
closes https://github.com/expressjs/discussions/issues/285

Note: this template was used in #290 and #289

PR #296, opened 2024-10-28
Add a section in the README for ADRs.

bjohansebas:
Related #291


UlisesGascon (approved):

.github/PULL_REQUEST_TEMPLATE/adr-template.md
```diff
@@ -0,0 +1,77 @@
+---
+name: ADR Template
+description: Template for Express Architecture Decision Records
+labels: ADR
+---
+
+# ADR template
+
+This is the base template that we use
```
wesleytodd, line 9:
```suggestion
This is the base template that we use. Sections are optional, if they are not applicable to your proposal please remove them.
```

.github/PULL_REQUEST_TEMPLATE/adr-template.md
```diff
@@ -0,0 +1,77 @@
+---
+name: ADR Template
```
wesleytodd, line 2: Since I was not aware of the ADR format, I wonder if we could use more commonly used terminology here? RFC is the one I find most common and I know @sheplu even mentioned that on our call. That said, "proposal" would also make sense to me, and I think is more in the loose spirit of what these can be.

wesleytodd (changes requested):
I put request changes, but I think it is not truly something I consider blocking. I am just worried folks will be like me and need to go look up what "ADR" means to understand.

.github/PULL_REQUEST_TEMPLATE/adr-template.md
UlisesGascon: RFC sounds great! :)

carpasse (approved):

bjohansebas:
I was thinking about this PR, and I don't think it's best to include all the content of the ADR as the message of the PR, as it would just duplicate the content that the PR would carry. Perhaps we should provide a message stating 'use the template located in X place' and add a section in the README educating about the ADRs. What do you think?

UlisesGascon:
I think that is good idea @bjohansebas, so we only make changes in one place too.

.github/PULL_REQUEST_TEMPLATE/adr-template.md
ctcpip: an RFC is something different and I think we should prefer the semantically correct term for what this represents.  `RFC !== ADR`

jonchurch:
I don't see the need for a PR template for this. 

Functionally, unless it is the default PR template (named `pull_request_template.md`), the only way to open a PR w/ this template is via querystrings specifying the template name and visiting that link. People create workarounds for that but they're clunky.

The value I do see here is adding an `ADR` label. We can do that via an action like

```yaml
name: ADR Labeler

on:
  pull_request:
    types: [opened, synchronize]
    paths:
      - 'docs/adr/**'

jobs:
  label-adr:
    runs-on: ubuntu-latest
    permissions:
      pull-requests: write
    
    steps:
      - uses: actions/github-script@v7
        with:
          script: |
            # Adding ADR label - if it already exists, this is a no-op
            github.rest.issues.addLabels({
              owner: context.repo.owner,
              repo: context.repo.repo,
              issue_number: context.issue.number,
              labels: ['ADR']
            })
```

bjohansebas:
I updated it to remove the template and fix the meetings section in the readme.

.github/PULL_REQUEST_TEMPLATE/adr-template.md
bjohansebas: Agree with Chris

README.md
```diff
@@ -5,13 +5,28 @@ This repository is used for generalized Express.js discussions.
...
 The Express.js Technical Committee (TC) meets online every other week, providing
 a live stream of the discussion and a recording of past meetings. Topics for the
 meetings are determined on an as-needed basis and may be technical or management
 of nature.
 
-Current meeting schedule: Wednesdays 23:30 UTC
+Current meeting schedule: 
+
+- TC Meeting: Monday 21:00 UTC (Every 2 weeks)
+- Working Session: Wednesday 21:00 UTC (Every 2 weeks)
```
wesleytodd, line 18: I wonder if we should just replace this with a link to the calendar or something? Either way, this seems fine to land as is now.

PR #300, opened 2024-11-02
Create adr-labeler workflow

jonchurch:
Related to https://github.com/expressjs/discussions/pull/296#issuecomment-2453123631

Will label any PR that touches the adr dir w/ `ADR` label

### DIFF
--- a/.github/workflows/adr-labeler.yml
+++ b/.github/workflows/adr-labeler.yml
@@ -1,0 +1,25 @@
+name: ADR Labeler
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+      - 'docs/adr/**'
+
+jobs:
+  label-adr:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          script: |
+            # Adding ADR label - if it already exists, this is a no-op
+            github.rest.issues.addLabels({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              labels: ['ADR']
+            })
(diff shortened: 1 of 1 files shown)


bjohansebas:
I like it, although I'm sure it will give an error like in https://github.com/expressjs/expressjs.com/issues/1553

jonchurch:
Right, default token perms are read on prs

I think it needs this update

```

permissions:
  pull-requests: write
  issues: write

```

bjohansebas:
I still think it wouldn't work, we tried that in https://github.com/expressjs/expressjs.com/pull/1557 and it didn't work. The only way we believe might work, although we haven't tested it yet, is by creating a token with the permissions as attempted in https://github.com/expressjs/expressjs.com/pull/1642.

jonchurch:
Ah, it's the permission model for forks mucking that up in the linked issue.

See the [docs for GH_TOKEN perms](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token), there's a note about this at the bottom.

I'll push an update to allow this to run based on PRs from forks.

jonchurch:
After looking into this, idk that there's a safe way to do this tbh. We can use `pull_request_target` to give forks the right permissions, but then they could possibly leak the token w/ those elevated perms. Same w/ the PAT approach, could trigger a workflow run, alter the workflow file in the PR, leak the token.

The original PR works as long as folks aren't opening PRs based on forks, but creating branches here in the repo.

TIL: the permission model of GHA is pretty wild

wesleytodd:
Can we change the name to RFC? I thought we had discussed that, but maybe with this new automation it is the time to do it?

bjohansebas:
due to permission issues, I don't think it's a good idea to add it, it will only create unnecessary noise about the workflow failing

bjohansebas:
I'm going to close it, but feel free to reopen it.

PR #306, opened 2024-11-14
feat(adr): establish a performance working group

wesleytodd:
This proposal would create a Performance Working Group. See the RFC content for details.

NOTE: this WG would operate under the STF proposal we have made. More details on that to come as we work through the process.

### DIFF
--- a/docs/adr/performance-wg.md
+++ b/docs/adr/performance-wg.md
@@ -63,7 +63,7 @@ purview:
 
 This funding does not mean the work can only be done by the owners, it just means that we are responsible for organizing
 and execuring on the deliverables. This is our first attempt at doing a program like this, so hopefully we can deliver
-on this well and show other orginacitions that investing in Open Source is worthwhile.
+on this well and show other organizations that investing in Open Source is worthwhile.
 
 ## References
 
--- a/docs/adr/performance-wg.md
+++ b/docs/adr/performance-wg.md
@@ -1,0 +1,71 @@
+# ADR [Number]: Performance Working Group
+
+## Status
+
+Proposed
+
+## Submitters
+
+- @wesleytodd
+
+## Decision Owners
+
+- @expressjs/express-tc
+
+## Context
+
+Express has traditionally taken a very lax approach to performance. Historically this has meant poor outcomes with
+regard to performance and lack of visibility around how changes impact performance. The goal of this proposed working
+group is to centralize discussion so the whole ecosystem of packages can benefit from targeted performance work. The
+scope of this WG would be all active packages in the `expressjs`, `pillarjs` and `jshttp` orgs with a special focus on
+the ones which are direct dependencies of `express` itself. 
+
+**Why do we need a working group?**
+
+The current state is we have many open PRs, in flight initiatives, and planned future work. With a loose approach we
+can easily have duplicated work, conflicting approaches, or focus on the wrong outcomes. By giving this WG scope over
+this we can centralize the discussion and plans so that we unlock better and faster work in the individual packages.
+
+## Decision
+
+Similar to the Security WG, we will setup a new repo (expressjs/perf-wg) and corresponding plan of action. The WG will
+be delegated responsibility for driving concensus around the current scope and future scope to be determined by the
+group once setup. This group will have three primary goals:
+
+1. Organize the effort. We want to make sure we are not rebuilding the same things or having the same disucssions in
+   many package repos at once.
+2. Provide and maintain tooling for all the packages/repos to produce reliable performance metrics and benchmarks. This
+   should include but is not limited to: workflows, infra and guidelines on types of benchmarks.
+3. Proritize and execute performance improvements across the packages. These do not need to be done *by* the members of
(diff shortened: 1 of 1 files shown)


bjohansebas (approved):
LGTM

andrehrferreira:
Hey guys, how are you?

I took some time off to take a year-end break, I recently performed some tests using my integration tool with the Node.js inspector, I'll leave the link to the test repository, https://github.com/andrehrferreira/express-benchmarks, the test is very simple, I uploaded an instance of express without any middleware and performed the monitoring using the inspector, then I used autocannon to trigger multiple requests in this application using the command `autocannon -c 100 -d 40 -p 10 localhost:3000`

As a result, my system creates a cpuprofile file that I analyze using the application https://discoveryjs.github.io/cpupro, I'll leave the profile file in the repository for deeper analysis, but basically, `37.5%` of the processing time was from internal express functions, with `13.8%` from response.js and 9% from the router, which in my opinion are the main ones bottlenecks of the current version, 7.3% in middleware/init.js, looking deeper the most worrying is in fact the send function for response, followed by the header and contentType, as I had already commented before in my point of view the main problem is in the use of Object.defineProperty, however I will continue analyzing other important points to be optimized

wesleytodd:
Hey @andrehrferreira, thanks for the post. These kinds of things are great, but for the project the larger problem is two fold:

1. Tracking performance as we change and improve the libraries (also taking into account runtime improvements)
2. Testing *real application* performance where the framework is impactful

It has been my experience that many engineers will spend a lot of time on optimizing some small subsystem that has little to no impact on the real world performance of their applications. My goal with setting up this working group is to level up everyone who wants to do perf work and to give ourselves the tools to do the *right analysis* for the right type of changes. Please keep finding things, and preferably open PRs with a change, a before and after benchmark result, and a description of how that impacts real applications (is it hot path, startup, etc), but also lets post that in the applicable repos instead of here where we are bootstraping the WG. Thanks!

---

Onto the topic of the WG, I am going to be just starting the work on this next week. I am going to update this proposal soon to include some more detailed goals/outcomes so we can be clear on what this is for.

wesleytodd:
I just pushed some updates to the proposal. Adding the TC for a review now that it has more details.

docs/adr/performance-wg.md
```diff
@@ -0,0 +1,71 @@
...
+  - Delivery Date: Dec 31, 2025
+  - Estimated Budget: €25,200
+2. Milestone 9: Improve performance for Express and critical dependencies
+  - Owner: @wesleytodd & @blakeembrey
+  - Delivery Date: Jun 30, 2026
+  - Estimated Budget: €46,200
+
+This funding does not mean the work can only be done by the owners, it just means that we are responsible for organizing
+and execuring on the deliverables. This is our first attempt at doing a program like this, so hopefully we can deliver
+on this well and show other orginacitions that investing in Open Source is worthwhile.
```
bjohansebas, line 66:
```suggestion
on this well and show other organizations that investing in Open Source is worthwhile.
```

docs/adr/performance-wg.md
```diff
@@ -0,0 +1,71 @@
+# ADR [Number]: Performance Working Group
```
bjohansebas, line 1: Don't forget to add the number before merging

docs/adr/performance-wg.md
```diff
@@ -0,0 +1,71 @@
...
+  - Owner: @wesleytodd
+  - Delivery Date: Dec 31, 2025
+  - Estimated Budget: €25,200
+2. Milestone 9: Improve performance for Express and critical dependencies
+  - Owner: @wesleytodd & @blakeembrey
+  - Delivery Date: Jun 30, 2026
+  - Estimated Budget: €46,200
+
+This funding does not mean the work can only be done by the owners, it just means that we are responsible for organizing
+and execuring on the deliverables. This is our first attempt at doing a program like this, so hopefully we can deliver
```
asminkarki012, line 65: "I noticed a typo in line 65: 'Execuring' should likely be 'Executing'."

wesleytodd:
```suggestion
and executing on the deliverables. This is our first attempt at doing a program like this, so hopefully we can deliver
```

Thanks @asminkarki012, an FYI for the future, if you find one you can use the "add a suggestion" option to make one like this, which I can then just apply right from the ui.

wesleytodd:
Just wanted to post this here since it came up on our WG call: https://github.com/openjs-foundation/infrastructure/issues/5

docs/adr/performance-wg.md
wesleytodd: I wonder if the number is really that helpful. Could we just use the PR number for this so we can updated it right away and have it correspond to the discussion we have here?

UlisesGascon (approved):
LGTM!

docs/adr/performance-wg.md
wesleytodd: I am going to do this, we can always amend later if we think it is a bad idea.

```suggestion
# ADR #306: Performance Working Group
```

bjohansebas:
Btw, i want to join this group.

PR #323, opened 2025-01-16
ADR: CommonJS and ESM decision

kjugi:
As discussed in our last meeting https://github.com/expressjs/discussions/issues/320

Summarizing the decision around ESM builds.


docs/adr/isomorphism-support.md
```diff
@@ -0,0 +1,77 @@
...
+
+During [working session](https://github.com/expressjs/discussions/issues/320) we have decided to not revisit, investigate or discuss this topic further. That means ESM exports won't be available for expressjs as well for pillarjs and jshttp packages.
+
+**What will be done?**
+
+Future issues can be closed with a link to this document.
+
+## Rationale
+
+CommonJS is the default node.js syntax. The JS world moved in the ESM direction as browsers consumed it well, and bundlers could make a tree-shake feature and dynamic imports. There is still a lot of baggage within the ESM itself and the way how we can use it.
```
UlisesGascon, line 48:
```suggestion
CommonJS is the default syntax in Node.js. While the JavaScript ecosystem has increasingly moved toward ESM due to its compatibility with browsers, enhanced tree-shaking capabilities, and support for dynamic imports, there are still complexities and challenges associated with ESM. 
```

docs/adr/isomorphism-support.md
```diff
@@ -0,0 +1,77 @@
...
+
+**What will be done?**
+
+Future issues can be closed with a link to this document.
+
+## Rationale
+
+CommonJS is the default node.js syntax. The JS world moved in the ESM direction as browsers consumed it well, and bundlers could make a tree-shake feature and dynamic imports. There is still a lot of baggage within the ESM itself and the way how we can use it. 
+
+To keep it short it's a whole new chapter to discuss and consider so it could use a lot of time and resources to make it properly. Don't get this wrong - it's not impossible tho! Most of our users will use the package in their project and pass it to the bundler which will produce the right format without any issues.
```
UlisesGascon, line 50:
```suggestion
Adopting ESM for our libraries would require a significant investment of time and resources to ensure proper implementation and long-term maintenance. While it is not impossible to achieve, it represents a considerable effort. Moreover, the majority of our users already utilize our libraries in their projects, relying on bundlers to handle the necessary transformations without issues.
```

UlisesGascon (commented):
Great work doing the consolidation @kjugi! Thanks :heart: 

I added some suggestions regarding tone and context, feel free to discus/accept/reject them ;)

bjohansebas (approved):
LGTM, With the changes proposed by Ulises

docs/adr/isomorphism-support.md
```diff
@@ -0,0 +1,77 @@
...
+
+**Are there any existing issues/discussions/pull requests related to this?**
+- https://github.com/pillarjs/router/issues/128
+- https://github.com/expressjs/discussions/issues/297
+- https://github.com/expressjs/express/discussions/6051
+- https://github.com/jshttp/cookie/issues/211
+
+## Decision
+
+During [working session](https://github.com/expressjs/discussions/issues/320) we have decided to not revisit, investigate or discuss this topic further. That means ESM exports won't be available for expressjs as well for pillarjs and jshttp packages.
```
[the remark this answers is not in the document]
ljharb: This change suggests that you might accept PRs making the change; I’d probably add a sentence like “this also means PRs making this change will not be accepted”

docs/adr/isomorphism-support.md
kjugi: Not sure if I get that part:
> fwiw treeshaking CJS is identical to ESM in capability

based [on this](https://webpack.js.org/guides/tree-shaking/#conclusion) and [this](https://web.dev/articles/commonjs-larger-bundles)

> in order to take advantage of tree shaking, you must...
> - Use ES2015 module syntax (i.e. import and export).

Meanwhile I will accept this suggestion and we can update it again later if needed.

docs/adr/isomorphism-support.md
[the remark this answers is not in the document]
kjugi: both suggestions are applied in a separate commit. I'm resolving this thread

kjugi:
Thank you @UlisesGascon for your great suggestions! It elevates this document to a new level. It's more mature now :v:
@ljharb Thank you too for your comments and attention to detail! 🥇

wesleytodd (approved):
I am good with this as it is more of a "tracking document", but it may end up being something end-users look at and I am not sure it fully satisfies the questions they might have. We should keep an eye out for feedback on if we should make a user facing statement on this at some point, maybe as a blog post geared more toward end users.

UlisesGascon:
> maybe as a blog post geared more toward end users

100% that is something that we can do. Who wants to take leadership? :)

docs/adr/isomorphism-support.md
```diff
@@ -0,0 +1,84 @@
...
+
+During the [working session](https://github.com/expressjs/discussions/issues/320), we had an in-depth discussion about this topic. After careful consideration, we concluded that we will not make a dedicated effort to export our libraries in the ESM format. Instead, we will continue exporting the libraries as we have done historically. 
+
+This decision is motivated by the lack of resources to maintain such an effort in the long term. It is also worth noting that Express is specifically designed to run with Node.js. While some of our libraries can be considered "isomorphic," this was unintended and can currently be classified as an "unofficial but functional feature." Consequently, our CI systems do not include browsers or other runtimes as part of their testing workflows.
+
+At present, our libraries function seamlessly in Node.js, supporting both CommonJS and ESM. Transitioning to support additional scenarios, such as direct ESM exports, would require significant changes to our CI systems and additional maintenance overhead.
+
+**TL;DR**: Dedicated ESM exports will not be available for Express.js, PillarJS, or JSHTTP packages. PR with such a change will not be accepted.
```
ctcpip, line 46:
```suggestion
**TL;DR**: Dedicated ESM exports will not be available for [expressjs](https://github.com/expressjs), [pillarjs](https://github.com/pillarjs), or [jshttp](https://github.com/jshttp) packages. PRs with such a change will not be accepted.
```

docs/adr/isomorphism-support.md
```diff
@@ -0,0 +1,84 @@
...
+- **Alternatives Considered:**
+  - Alternative 1: Add ESM export to our libraries. CommonJS format is accepted by all most popular bundlers.
+- **Pros and Cons**: Outline the pros and cons of the chosen solution.
+- **Why is this decision the best option?** Time and energy can be shifted to other topics.
+
+## Consequences
+
+- **Positive Impact**: It does not require to support another set of tools and one more major (or at least big) release.
+- **Negative Impact**:
+  - Packages can't be used in deno projects and potentially in other future runtime engines for JavaScript that decide to not support commonjs. That can be a potential user miss
```
ctcpip, line 67: this is not true though, right?   deno can indeed use cjs

ljharb: can it? i thought jsr packages could be consumed by cjs but deno still can only consume esm packages

bjohansebas: the premise of Deno was that it was only ESM, although I would have to verify it. Now you've made me doubt.

ctcpip: https://deno.com/blog/v2.0-release-candidate#improved-commonjs-support

bjohansebas: Okay, they do support it, so it’s probably fine to say that only Deno 1 doesn’t support it, right?

ctcpip: I think we can just remove this line. or rather, change it to reflect that packages may or may not work in browser environments

bjohansebas: +1 'may not work in browser environments.'

docs/adr/isomorphism-support.md
```diff
@@ -0,0 +1,84 @@
...
+
+**Are there any existing issues/discussions/pull requests related to this?**
+- https://github.com/pillarjs/router/issues/128
+- https://github.com/expressjs/discussions/issues/297
+- https://github.com/expressjs/express/discussions/6051
+- https://github.com/jshttp/cookie/issues/211
+
+## Decision
+
+During the [working session](https://github.com/expressjs/discussions/issues/320), we had an in-depth discussion about this topic. After careful consideration, we concluded that we will not make a dedicated effort to export our libraries in the ESM format. Instead, we will continue exporting the libraries as we have done historically.
```
blakeembrey, line 40: I would love to expose ESM and did so historically, I presume I should not since it’d be a dedicated effort?

docs/adr/isomorphism-support.md
```diff
@@ -0,0 +1,84 @@
...
+- https://github.com/pillarjs/router/issues/128
+- https://github.com/expressjs/discussions/issues/297
+- https://github.com/expressjs/express/discussions/6051
+- https://github.com/jshttp/cookie/issues/211
+
+## Decision
+
+During the [working session](https://github.com/expressjs/discussions/issues/320), we had an in-depth discussion about this topic. After careful consideration, we concluded that we will not make a dedicated effort to export our libraries in the ESM format. Instead, we will continue exporting the libraries as we have done historically. 
+
+This decision is motivated by the lack of resources to maintain such an effort in the long term. It is also worth noting that Express is specifically designed to run with Node.js. While some of our libraries can be considered "isomorphic," this was unintended and can currently be classified as an "unofficial but functional feature." Consequently, our CI systems do not include browsers or other runtimes as part of their testing workflows.
```
blakeembrey, line 42: Being isomorphic has been intentional for the packages I maintain, and I do make a dedicated effort to keep it isomorphic. This includes edge environments that are plain JavaScript. The lack of CI for these environments aren’t connected to their support.

docs/adr/isomorphism-support.md
```diff
@@ -0,0 +1,84 @@
...
+  - Alternative 1: Add ESM export to our libraries. CommonJS format is accepted by all most popular bundlers.
+- **Pros and Cons**: Outline the pros and cons of the chosen solution.
+- **Why is this decision the best option?** Time and energy can be shifted to other topics.
+
+## Consequences
+
+- **Positive Impact**: It does not require to support another set of tools and one more major (or at least big) release.
+- **Negative Impact**:
+  - Packages can't be used in deno projects and potentially in other future runtime engines for JavaScript that decide to not support commonjs. That can be a potential user miss
+  - OSS library authors that use our packages in ESM native libs might suffer from a lack of support
```
blakeembrey, line 68: There is a greater negative impact which is that other users fork the packages to ESM and those packages lack security updates over time which leaves users in a worse position.

docs/adr/isomorphism-support.md
ljharb: I don't agree with that last statement; something isn't actually supported unless it's actually tested.

blakeembrey: That’s a reasonable take, but doesn’t preclude that I do support them. Some things are unreasonable to have direct tests. There’s a reason we often only test each major node release instead of every single version, and we don’t claim those aren’t supported. It’s the same that I don’t test directly in CI for Cloudflare Workers (honestly not sure how I would tbh) but I do support them, by virtue of knowing it’s V8 and disallowing native node APIs.

blakeembrey: TL;DR is that I don’t think this assertion belongs in the doc, something being isomorphic may be related but is not connected to ESM support. I don’t want to give an impression that some of the packages in these organizations are not designed to support other JavaScript environments, regardless of ESM support, since it could be referred to as a way to drop existing supported environments without a major.

ctcpip: I agree, we should decouple isomorphism from module system compatibility.  The root of this was that ESM exports would be justified for packages for which we wanted to support isomorphism.  In other words, if the package was not intended to be isomorphic, there is perceived little value in exporting ESM.

I also agree that we cannot (and should not attempt) to test against every possible environment.  However, isomorphic packages should get _some_ level of testing outside of node, and this should hopefully not be prohibitively difficult.

We have already [run into situations](https://github.com/pillarjs/router/issues/128#issuecomment-2435666417) where we are not certain if a package is properly working in other environments, in part due to lack of isomorphism in the test suite.

ljharb: Packages can be perfectly isomorphic without ESM, though - it's only if someone insists on not having a build process that the module format makes a difference.

ctcpip: Absolutely.  I just think it's easier to justify if e.g. targeting the browser, not that it's necessarily required.

docs/adr/isomorphism-support.md
ctcpip: Which packages export ESM today?

blakeembrey: Historically I did it for path-to-regexp but removed it in the latest major to align with the TC, it wasn’t any additional overhead to maintain.

docs/adr/isomorphism-support.md
wesleytodd: Do we have a list of examples where the docs or tests indicate browser support? Would just be good to have a clear list of the places we have stated a policy so we ensure any new work keeps in line with that.

I fully agree we should decouple ESM (this doc) from browser support, but I think it is reasonable to mention that ESM provides no benefits to server only libraries and that it informed our decision. Maybe we could make a small change like removing the "never intended" for softer language around that?

```suggestion
This decision is motivated by the lack of resources to maintain such an effort in the long term. It is also worth noting that Express was specifically designed to run with Node.js. While some of our libraries can run in other runtimes, this can currently be classified as an "unofficial and untested feature." Consequently, our CI systems do not include browsers or other runtimes as part of their testing workflows.
```

blakeembrey: At a risk of derailing the topic I'll reply.

> However, isomorphic packages should get some level of testing outside of node, and this should hopefully not be prohibitively difficult.

Absolutely, but the overhead isn't valuable for some packages. In `path-to-regexp`, by not allowing `node` imports and testing on a minimum node version for JavaScript coverage I'm 99% of the way there. The 1% extra hasn't been worth the overhead (time and money) of e.g. an iOS Safari browser test suite hasn't been worth it, but if someone wanted to donate those things I wouldn't say no.

This wouldn't apply to something more complex that had code paths that change depending on their environment, but in this case it's just plain JavaScript with no dependencies.

> Do we have a list of examples where the docs or tests indicate browser support?

No, but I'm in favor of standardizing some way to indicate it.

 > I think it is reasonable to mention that ESM provides no benefits to server only libraries and that it informed our decision.

Dead code elimination can still be an issue for server environments, e.g. Cloudflare Workers has a 3MB limit. IIRC AWS Lambda and other platforms have similar requirements. I know dead code is meant to work with CommonJS but I haven't had much luck and ESM is the only time it actually works for me.

ljharb: There's no technical difference between what static analysis is doable for CJS and ESM, so while the reality of tool support may differ, there's no actual justification for it.

blakeembrey: Maybe but sadly we’re stuck with the tooling we have today, not the ideal world, so I thought it worth mentioning this isn’t just a browser concern.

Ref: https://github.com/evanw/esbuild/blob/main/docs/architecture.md#es6-linking

> This is possible with ES6 modules but not with CommonJS because ES6 imports are bound at compile time while CommonJS imports are bound at run time.

Just to be clear, I’m not at all advocating for ESM here, I’m only providing context on why it shouldn’t be dismissed as “it’s a browser thing”.

ljharb: ESM imports can be static or dynamic; so can CJS requires - since tooling has to account for both, there's literally no technical difference between them. Nothing is possible for ESM that's not possible for CJS if you're talking about a build tool.

blakeembrey: That sounds like something to take up with the tools, I’m just pointing out that common build tools (both Webpack and esbuild) require ESM for tree shaking, and these are common tools to bundle for server side environments that may have size limits (including ones that run node).

Again, don’t take this as arguing for ESM support in this thread, but to argue against that it’s only relevant for browsers.

ljharb: I completely agree :-) i just think it's important to point out that it's a failing in the tools, not an inherent difference in the module format.

docs/adr/isomorphism-support.md
kjugi: That's a valid point! Added, thanks!

docs/adr/isomorphism-support.md
kjugi: Ok, I have applied some changes. Please verify. Also, I think I'm parking the "isomorphic" topic completely. As mentioned in the comments, let's focus on export type here. I can start another doc about isomorphic and its status right after this one if needed.

docs/adr/isomorphism-support.md
kjugi: Do you think I should include this fact from the past? I've been thinking about how to put this together now and I'm not sure.
The document should state the last known status. We can list exceptions now (if any) and explain this fact or update the document later when we agree on which one we want to change. There is a Changelog on the bottom so it should be easy to mark that out in the future.

docs/adr/isomorphism-support.md
kjugi: Updated!

UlisesGascon (approved):

docs/adr/isomorphism-support.md
bjohansebas: We would need to change the phrase 'historically,' as it gives the impression that it was never done, but as Blake explains, it was done in some package.

docs/adr/esm-support.md
```diff
@@ -0,0 +1,85 @@
...
+
+This decision is motivated by the lack of resources to maintain such an effort in the long term. It is also worth noting that Express was specifically designed to run with Node.js. While some of our libraries can run in other runtimes, this can currently be classified as an "unofficial and untested feature." Consequently, our CI systems do not include other runtimes as part of their testing workflows.
+
+At present, our libraries function seamlessly in Node.js, supporting both CommonJS and ESM. Transitioning to support additional scenarios, such as direct ESM exports, would require significant changes to our CI systems and additional maintenance overhead.
+
+**TL;DR**: Dedicated ESM exports will not be available for [expressjs](https://github.com/expressjs), [pillarjs](https://github.com/pillarjs), or [jshttp](https://github.com/jshttp) packages. PRs with such a change will not be accepted.
```
kjugi, line 46: This part needs polishing. Soften the language and consider exporting ESM when it requires little effort. Include libraries maintained by @blakeembrey, which now include ESM export. It should be clear that our CI tests do not cover this for today.
"CommonJS is a requirement, but ESM is not"

samuelstroschein:
Chipping in from an "I want to use express dependencies but need ESM" perspective, see https://github.com/pillarjs/path-to-regexp/issues/347. 

If the maintenance effort of libraries like `path-to-regexp` to distribute CJS and ESM does not lead to additional overhead (@blakeembrey in https://github.com/expressjs/discussions/pull/323#discussion_r1929203580), why should the desire for express js to consume dependencies as CJS negatively impact the dependencies and web ecosystem?

wesleytodd:
> I assume that experience will be universal across express js dependencies.

This was a main point of our discussion. It is not clear to me that it should be universal. The one thing we all agreed on is that we MUST ship CJS, but there seems to be no reason to limit packages which want to ship ESM from doing so. That was the main feedback on this doc from our meeting yesterday IIUC.

ljharb:
There are inherent costs and dangers in shipping a dual-format package - it exacerbates the likelihood of duplicated (and thus out of sync) state; it's unlikely both formats are being tested (whether they're hand-written, increasing the likelihood of deviations, or transpiled, increasing the likelihood of transformation bugs); it increases the package size (which seems to be important to the same folks that want native ESM).

Also, node can import CJS, and every build tool can translate it to ESM just fine if needed, so I'm still not clear on the value.

samuelstroschein:
@ljharb hope to share some light on the necessity of ESM. 

Ofc you can make the decision to ignore the (minor) group that needs ESM. That's completely OK. Everything is a trade-off. 

> , so I'm still not clear on the value.

I can't adopt `path-to-regexp` if it's not ESM because tree-shaking doesn't work https://github.com/pillarjs/path-to-regexp/issues/347 and import errors occur https://github.com/pillarjs/path-to-regexp/issues/346. 

 >  it increases the package size (which seems to be important to the same folks that want native ESM).

ESM people care about tree-shaking, not the package size.

wesleytodd:
Yeah, I agree that I continue to be unclear on the value. That said, we did discuss on the call yesterday that we should expect them to be tested if we are shipping dual mode packages.

ljharb:
@samuelstroschein right - i'm saying that CJS treeshakes the same as ESM, and if there's a specific tool you're using that doesn't support it, i'd encourage you to advocate they do so.

wesleytodd:
I would argue we should not cater to people who setup broken tooling or choose unrealistic and pointless goals. If someone comes with proof that there is some technical barrier to efficiently using CJS in a real world use case, I would absolutely love to help, but that is not what has happened in any of these threads (see my question in the one linked about treeshaking).

bjohansebas:
Btw, the development version of Multer (https://github.com/expressjs/multer/pull/399) is only for ESM, which means that in that development version, it should be converted to CommonJS.

docs/adr/esm-support.md
```diff
@@ -0,0 +1,85 @@
...
+
+**What will be done?**
+
+Future issues can be closed with a link to this document.
+
+## Rationale
+
+CommonJS is the default syntax in Node.js. While the JavaScript ecosystem has increasingly moved toward ESM due to its compatibility with browsers, enhanced tree-shaking capabilities (coming from bundler tools), and support for dynamic imports, there are still complexities and challenges associated with ESM.
```
slagiewka, line 54: If there is any document that supports the "default syntax" statement, it would be beneficial to link it. 

It seems to me that in Node, CJS is the _de facto_ standard because of its history but it's not set in stone. Especially with syntax detection enabled by default in all LTS versions after April 2025 (when 18 is gone) - [Node 20.19](https://nodejs.org/en/blog/release/v20.19.0#module-syntax-detection-is-now-enabled-by-default) has it enabled.

slagiewka (commented):
For going ESM-only, the elephant in the room is TypeScript. Users wanting to run `require(esm)` (because their output from tsc is CJS) will need to either set `--module nodenext` when on Typescript 5.8 OR wait for 5.9 to use `--module node20`.

More on that here: https://github.com/microsoft/TypeScript/pull/60761

If this doc would consider next major releases that target Node 20+ the summary would be simple: choosing ESM over CJS is a matter of preference. Publishing dual package makes no sense.
However, it can also be a matter of vision. If the express ecosystem wants to be considered modern, it would use modern standards as long as it allows the stability it's renown for.

docs/adr/esm-support.md
ljharb: node’s own docs, the default value of the “type” field, the default value of node’s input-type arg… does it really require evidence when it’s so overwhelmingly the case?

slagiewka: This says absolutely nothing about what the project considers the default. But you know why? Let's go down the rabbit hole.

Back when the toggle was [first added](https://github.com/nodejs/node/pull/37162) it already had `mjs` put first into the Markdown and `cjs` below it. It still is the case to this day.

In author's own words:
> Enable code example using both modern ESM syntax and legacy CJS syntax.

Want more examples?
https://github.com/nodejs/node/pull/37170
> This is done to put emphasis on the ESM and Promise model.

Commit message says:
> Convert examples to esm to help promote that pattern with users

Now, this only shows _intent_ from authors. But is much better case (in my opinion) than the toggle.

I'd rather this decision was based on evidence and not coincidence that some toggle defaults to this or that state. And this could as well be a "bug" to fix with the docs page.

ljharb: The default is whatever happens without configuration with .js files, which is always CJS.

it is not a bug, it is an intentional decision that is unlikely to ever change because it would break too much of the ecosystem.

slagiewka: > The default is whatever happens without configuration with .js files, which is always CJS.

Let's do a test.

```sh
❯ mkdir testesm && cd testesm
❯ echo 'import {version} from "node:process"; console.log(version);' > test.js
❯ node test.js
```

Which will fail on node 18:
```
❯ /opt/homebrew/Cellar/node@18/18.20.8/bin/node test.js
(node:47428) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/Users/user/Projects/testesm/test.js:1
import {version} from "node:process"; console.log(version);
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:76:18)
    at wrapSafe (node:internal/modules/cjs/loader:1283:20)
    at Module._compile (node:internal/modules/cjs/loader:1328:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Module._load (node:internal/modules/cjs/loader:1019:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:128:12)
    at node:internal/main/run_main_module:28:49

Node.js v18.20.8
```

But will have ABSOLUTELY NO ISSUE on 20+.

```
❯ node test.js
v20.19.0
```

Therefore it's not **_always_** CJS.

Let's extend the test further:
```sh
❯ echo '{"name": "testesm"}' > package.json
```

```
❯ node test.js
v20.19.0
(node:48221) [MODULE_TYPELESS_PACKAGE_JSON] Warning: Module type of file:///Users/user/Projects/testesm/test.js is not specified and it doesn't parse as CommonJS.
Reparsing as ES module because module syntax was detected. This incurs a performance overhead.
To eliminate this warning, add "type": "module" to /Users/user/Projects/testesm/package.json.
(Use `node --trace-warnings ...` to show where the warning was created)
```

Even with _poorly packaged_ code your statement does not hold true and node.js does not default to CJS. Nothing breaks in the ecosystem. 

The only actual evidence of "default" would be here: https://nodejs.org/docs/latest/api/packages.html#introduction

> In particular, package authors should always include the ["type"](https://nodejs.org/docs/latest/api/packages.html#type) field in their package.json files, even in packages where all sources are CommonJS. Being explicit about the type of the package will future-proof the package in case the default type of Node.js ever changes

And the `type` documentation is also misleading, as it states:
> import statements of .js files are treated as ES modules if the nearest parent package.json contains "type": "module"

Which is not true, because omitting `type` does not mean `commonjs` by default. 
```
❯ echo '{"name": "testesm", "type": "commonjs"}' > package.json
```
which breaks ESM inside `.js` files.

This shows that `no type != commonjs`. If there's any "default", it's "syntax detection".

ljharb: You're right, that's fair that the default is now "syntax detection" for files. However, for the REPL, it's still CJS, and i think this sentence remains correct.

docs/adr/esm-support.md
```diff
@@ -0,0 +1,85 @@
...
+Adopting ESM for our libraries would require a significant investment of time and resources to ensure proper implementation and long-term maintenance. While it is not impossible to achieve, it represents a considerable effort. Moreover, the majority of our users already utilize our libraries in their projects, relying on bundlers to handle the necessary transformations without issues.
+
+- **Alternatives Considered:**
+  - Alternative 1: Add ESM export to our libraries. CommonJS format is accepted by all most popular bundlers.
+- **Why is this decision the best option?** Time and energy can be shifted to other topics.
+
+## Consequences
+
+- **Positive Impact**: It does not require to support another set of tools and one more major (or at least big) release.
+- **Negative Impact**:
```
slagiewka, line 65: I think it would be justified to acknowledge that the decision means that Express libraries will not be published to JSR (or any other registry that will be ESM-only).

docs/adr/esm-support.md
wesleytodd: @slagiewka, If you disagree with the wording, please suggest an alternative and we can decide if it is a better fit. Otherwise I think we can resolve this thread, because I don't think nitpicking this wording choice is worth it if the goal is not to pick other words. Especially since these were chosen intentionally and are technically correct enough.

docs/adr/esm-support.md
```diff
@@ -0,0 +1,85 @@
...
+  - Potential community library fork (to make it ESM-friendly) might lack security updates over time
+  - OSS library authors that use our packages in ESM native libs might suffer from a lack of support
+- **Mitigations**: Potential decision update for selected libraries (not specified yet) and exposing both types (CJS and ESM)
+
+## References
+
+Support for commonjs imports in ESM code is available in the node. Described in docs:
+- https://nodejs.org/api/esm.html#interoperability-with-commonjs
+
+Support for ESM modules imports in commonjs is available since node v20 behind the experimental flag and node v23 without a flag. Docs:
```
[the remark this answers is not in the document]
wesleytodd:
```suggestion
Support for ESM modules imports in commonjs is available in modern Node.js versions. Docs:
```

Lets just drop the moving target and reference the upstream docs.

docs/adr/esm-support.md
ctcpip: I suggest that's somewhat out of scope. Yes, it would prevent publishing to certain registries, but which registries we use should not be coupled to this ADR.  We can't publish to maven either, but we don't need to point that out.

wesleytodd: Agreed, calling this resolved.

docs/adr/esm-support.md
slagiewka: > Especially since these were chosen intentionally and are technically correct enough.

I'm yet to see it be proven technically correct. There is no default standard  in node.js and there is no suggested syntax for libraries to use. Granted, the state of lesser complexity is fairly new.

> nitpicking this wording choice

This is the leading statement that serves as the rationale for the decision. And at the time of writing is already hard to sustain. It may skew the decision towards suboptimal choice. However, I sense that everyone's mind is already set anyway.


> If you disagree with the wording, please suggest an alternative

Given the above, there are a couple of options:

1. Remove what can't be supported
```suggestion
While the JavaScript ecosystem has increasingly moved toward ESM due to its compatibility with browsers, enhanced tree-shaking capabilities (coming from bundler tools), and support for dynamic imports, there are still complexities and challenges associated with ESM. 
```
2. Try describing node's limbo.
```suggestion
CommonJS is still the dominating syntax in Node.js userland. While the JavaScript ecosystem has increasingly moved toward ESM due to its compatibility with browsers, enhanced tree-shaking capabilities (coming from bundler tools), and support for dynamic imports, there are still complexities and challenges associated with ESM. Node.js has only recently gained enough support for libraries to transition to ESM without requiring users to move away from CommonJS.
```

docs/adr/esm-support.md
slagiewka: I'm fine with this not being mentioned, but I would expect such thing to be at least **_acknowledged_**. That's just due diligence for making a technical decision, which this document is.

>  We can't publish to maven either, but we don't need to point that out.

This is _reductio ad absurdum_. We're talking about a well established (to say the least) JavaScript library/ecosystem and a registry/registries that are serving JavaScript in its TC39-defined standard.

docs/adr/esm-support.md
ljharb: As I said, it's the default syntax in the repl, which should be sufficient to establish that it's the default syntax for node.

Also, node 18 isn 't EOL quite yet, and it will be years after that that libraries can painlessly transition to ESM.

ctcpip:
[This proposed ADR](https://github.com/expressjs/discussions/pull/365/files) carves out the question of environment support, so that this ADR can focus on CJS vs ESM rather than runtime/env support.

ctcpip (changes requested):
blocking for now until we resolve all issues, and, ideally, settle on the language for a [related ADR](https://github.com/expressjs/discussions/pull/365/files), which has, in a sense, a broader scope, and should allow us to clean up some of the language here

dynst:
A switch actually requires more than just dropping support for old Node.js versions.

This block of code in the `View` constructor is currently dynamically, and synchronously, calling `require(arbitrarystring)`, but in ESM, that kind of arbitrary import would have to be part of an `async` function, because it would have to use `await import(...)` syntax instead.

https://github.com/expressjs/express/blob/54af593b739ea44674e4a445efa15b8024f093da/lib/view.js#L75-L88

jorenbroekema:
There's a fair amount of use cases for ESM builds of libraries such as `path-to-regexp` which do not follow what seems to be a primary premise of this ADR, which is that the tools are designed for NodeJS use cases.

Perhaps that's true also for `path-to-regexp`, as it was probably created with a NodeJS environment in mind, but nothing in that package is specifically NodeJS, it's quite agnostic to its environment.
Fortunately there is a browser API (URLPattern) inspired by it for browser usage nowadays, and with 90% global usage browser support as of writing this as well as a polyfill available. 

That said, in browser-land, there's a lot of packages still making use of `path-to-regexp`, usually v6 and below because those versions still published as dual ESM/CJS. There are vulnerabilities (CVE-400 comes to mind) that can only be resolved by migrating to v8 which is CJS-only.

"Why not just use CJS, just bundle..."

I know that in most ecosystems, having a CJS-to-ESM conversion happening in your local dev and build tools is common (Vite being the most popular at this time), that said, there are plenty of ecosystems where the focus is on "browser-compatible out of the box". For example, CDNs will often mirror NPM registry and only handle bare import specifier resolution (or not, and expect you to deal with it via importmap, JSPM is an example of such a CDN), and nothing else, meaning that CJS-only packages are not usable from such CDNs. Especially in corporate environments where microfrontends and heavy CDN usage is quite common, this poses a significant challenge. This is just one example, I just happen to be quite active in frontend environments where a CJS-to-ESM conversion isn't a given, or is actively avoided for toolchain-minimalism or performance-related reasons.

Do what you will with that perspective... just wanted to make sure this is expressed and heard, and with some luck, considered in the decision making around this.

docs/adr/323-esm-support.md
```diff
@@ -0,0 +1,85 @@
...
+Adopting ESM for our libraries would require a significant investment of time and resources to ensure proper implementation and long-term maintenance. While it is not impossible to achieve, it represents a considerable effort. Moreover, the majority of our users already utilize our libraries in their projects, relying on bundlers to handle the necessary transformations without issues.
+
+- **Alternatives Considered:**
+  - Alternative 1: Add ESM export to our libraries. CommonJS format is accepted by all most popular bundlers.
+- **Why is this decision the best option?** Time and energy can be shifted to other topics.
+
+## Consequences
+
+- **Positive Impact**: It does not require to support another set of tools and one more major (or at least big) release.
+- **Negative Impact**:
```
jorenbroekema: I want to add to this discussion that most of the libraries in https://pillarjs.github.io/ are not like ExpressJS in that they are quite NodeJS-specific. Most of these libs can and are actively used in browser-based projects/libraries. It seems odd to me that this ADR says it's mostly about express and projects that are specifically made for NodeJS, but then includes pillarjs which is imo very different and much more agnostic to use cases outside of NodeJS, browser being the main one.

docs/adr/esm-support.md
jorenbroekema: Just making sure this is visible here, https://github.com/pillarjs/path-to-regexp which is maintained by @blakeembrey  does not support ESM in its exports. This was removed in v7. 

I don't understand why pillarjs is included in the scope of this ADR, when there is a very clear use case for this to still publish ESM for out of the box browser compatibility. These libraries are way more often completely agnostic to NodeJS, so the line of logic of this ADR that we don't need to support ESM for packages that are specifically designed for NodeJS, does not apply to these libraries.

docs/adr/esm-support.md
jorenbroekema: @ljharb you're cherry picking things about NodeJS that still use CJS as a default due to historical reasons and not to cause needless breaking changes, as a way to support your argument that "CJS is the default syntax".

https://nodejs.org/en the homepage example uses ESM.
The NodeJS Docs for [CJS](https://nodejs.org/docs/latest/api/modules.html) and [ESM](https://nodejs.org/docs/latest/api/esm.html) respectively call:
- CJS: the **original** way to package JavaScript code for Node.js.
- ESM: the **official** standard format to package JavaScript code for reuse

There have been efforts to move to ESM as the defacto standard for many years now by the NodeJS ecosystem which is abundantly clear from Node's own developments towards this format.

What I'm seeing here is someone that is desperately picking at straws to support their anti-ESM crusade. I can only guess what your reasoning behind that is, I think most people in the NPM ecosystem are aware of this by now, but I fear the authors of this ADR are not..

jonkoops: > As I said, it's the default syntax in the repl, which should be sufficient to establish that it's the default syntax for node.

No, it means that it is the default syntax for the REPL, that is in no way reflective of the default behavior of imports, which now clearly default to module detection. And even if the REPL defaults to CommonJS, you can still import packages authored in ESM from it.

docs/adr/esm-support.md
ctcpip: I think we can just remove this paragraph entirely. This earlier statement captures reality:

> we will not make a dedicated effort to export our libraries in the ESM format

that doesn't mean that there is a blanket ban on ESM exports, it just means that as a project as a whole, we are not going down the path of making explicit efforts to go through all the packages and convert or add ESM exports

docs/adr/esm-support.md
ctcpip: Please be mindful of our [Code of Conduct](https://github.com/openjs-foundation/cross-project-council/blob/main/CODE_OF_CONDUCT.md#contributor-covenant-30-code-of-conduct).

Specifically:

1. Respecting the **purpose of our community**, our activities, and our ways of gathering.
2. Engaging **kindly and honestly** with others.
3. Respecting **different viewpoints** and experiences.
4. **Taking responsibility** for our actions and contributions.
5. Gracefully giving and accepting **constructive feedback**.
6. Committing to **repairing harm** when it occurs.
7. Behaving in other ways that promote and sustain the **well-being of our community**.

docs/adr/esm-support.md
```diff
@@ -0,0 +1,85 @@
...
+- https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require
+
+## Changelog
+
+Track changes or updates to this ADR over time. Include the date, author, and a brief description of each change.
+
+- **[2025-01-15]**: [@kjugi] - document init
```
ctcpip: I don't think we should track changes from before the ADR was approved and merged

docs/adr/esm-support.md
ctcpip: The disputed text wasn't doing any heavy lifting in this doc and therefore doesn't warrant further discussion.  It has been removed.

ctcpip (approved):

docs/adr/isomorphism-support.md
ctcpip: Language has been updated to reflect reality, particularly that repo captains can decide what to do with packages they are responsible for.

ctcpip:
Locking so we don't lose conversation in zombie PRs.  Please create a new issue or PR for further discussion/iteration if needed.

PR #346, opened 2025-03-10
fix: update meeting schedules to reflect new time adjustments

bjohansebas:
Once again, time changes according to my calendar


UlisesGascon:
I think the changes were expected as the US has started the Daylight Saving Time ([ref](https://www.timeanddate.com/time/change/usa))

bjohansebas:
Yep, but we still need to fix the meeting time, right?

wesleytodd (approved):
According to my calendar and monday's meeting they do need to change. Honestly seems like we need to make a patch to the meet tool or update our [template](https://github.com/expressjs/discussions/blob/master/.github/ISSUE_TEMPLATE/meeting.md?plain=1#L18) to account for DST? We should land this though, to fix it for now.

PR #352, opened 2025-03-28
feat(ADR): LTS Strategy

wesleytodd:
A proposal for an LTS strategy with committed dates and next steps. Please see the goals/non-goals for this ADR.

I attempted to copy/paste without edits from all the sources we had. I admit there was a lot of it and so if I missed something please help and link to it so I can get it included.

### DIFF
--- a/docs/adr/lts-strategy.md
+++ b/docs/adr/lts-strategy.md
@@ -128,15 +128,15 @@ For the existing release lines, we will set the following phase dates:
 
 | Major | CURRENT | ACTIVE | MAINTENANCE | EOL |
 | ----- | ------- | ------ | ----------- | --- |
-| 4.x   |         |        | 2025-04-01 | *no sooner than 2026-10-01 |
-| 5.x   | 2024-09-11 | 2025-03-31 | **TBD | **TBD |
-| 6.x   | ***TBD | | | |
+| 4.x   |         |        | 2025-04-01 | <sup>[1]</sup> no sooner than 2026-10-01 |
+| 5.x   | 2024-09-11 | 2025-03-31 | <sup>[2]</sup>TBD | <sup>[2]</sup>TBD |
+| 6.x   | <sup>[3]</sup>TBD | | | |
 
-*: v4 is a special case, and we may extend MAINTENENCE support. This date is called out to give users confidence we will
+1. v4 is a special case, and we may extend MAINTENENCE support. This date is called out to give users confidence we will
    not end-of-life *earlier* than this date.
-**: v5 MAINTENENCE and EOL dates are determined by when v6 is released. We will update them when v6 release dates are
+2. v5 MAINTENENCE and EOL dates are determined by when v6 is released. We will update them when v6 release dates are
     committed to. 
-***: v6 work has not started yet. The earliers we could choose to release by this proposal is 2026-01-01. We will update
+3. v6 work has not started yet. The earliers we could choose to release by this proposal is 2026-01-01. We will update
      these dates when we commit to them.
 
 ### Documentation
--- a/docs/adr/lts-strategy.md
+++ b/docs/adr/lts-strategy.md
@@ -77,7 +77,7 @@ requirements:
     tagged `latest` on npm for a minimum of 12 months.
   - `MAINTENANCE`: When a new major version becomes `ACTIVE`, the previous major version enters `MAINTENANCE` for 12 months.
 - After the `MAINTENANCE` time has ended the major version is considered `EOL` and is unsupported.
-- During the `ACTIVE` period, a new major version may be released but *not* until at least 6 months have passed
+- During the `ACTIVE` period, a new major version may be released but *not* until at least 12 months have passed
   since the `CURRENT` version became `ACTIVE`.
 - Users are required to follow the head/latest of each major release line for support with all packages (`express`,
   dependencies, middleware, & tools/other)
@@ -136,9 +136,14 @@ For the existing release lines, we will set the following phase dates:
 
 | Major | CURRENT | ACTIVE | MAINTENANCE | EOL |
 | ----- | ------- | ------ | ----------- | --- |
-| 4.x   |         |        | 2025-04-01 | 2026-10-01 or later |
-| 5.x   | 2024-09-11 | 2025-03-31 | 2026-04-01 | 2027-04-01 |
-| 6.x   | TBD after 2025-10-01 | | | |
+| 4.x   |         |        | 2025-04-01 | *no sooner than 2026-10-01 |
+| 5.x   | 2024-09-11 | 2025-03-31 | **no sooner than 2026-04-01 | **no sooner than 2027-04-01 |
+| 6.x   | ***no sooner than 2026-01-01 | | | |
+
+*: v4 is a special case, and we may extend MAINTENENCE support
+**: v5 MAINTENENCE and EOL dates are determined by when v6 is released, these dates reflect the earliest datest if we
+were to ship v6 on 2025-10-01
+***: v6 work has not officially started yet, this is simply the earliest date we can ship based on our proposed policy
 
 ### Documentation
 
--- a/docs/adr/lts-strategy.md
+++ b/docs/adr/lts-strategy.md
@@ -86,7 +86,7 @@ requirements:
 - Dropping old Node.js versions *alone* is not enough of a breaking change to release a new major version. Applies to
   `express`, dependencies, middleware, & tools/other.
 - `express` version 4.x is considered a special case. It will receive a longer `MAINTENENCE` phase of 18 months or until
-  the TC agrees we are save cutting support.
+  the TC agrees we are safe cutting support.
 - `express` dependency packages will follow the same timeline and support dates for the `express` version which they are
   included with.
 - Middleware packages should follow the timeline and support dates for the `express` versions they are compatible, but
--- a/docs/adr/lts-strategy.md
+++ b/docs/adr/lts-strategy.md
@@ -1,0 +1,176 @@
+# ADR [Number]: LTS Strategy and Support Schedule
+
+## Status
+
+Proposed
+
+## Submitters
+
+- @wesleytodd
+
+## Decision Owners
+
+- @expressjs/express-tc
+
+## Context
+
+Conversation about our LTS strategy, support dates for our major release lines, and how we should message this content
+has been spread out across multiple issues, repos, slack channels, and meetings. We have landed a few starter documents,
+and generally have agreed on our goals/strategy but we have not officially documented and centralized that. This ADR is
+both gathering all that context to one place as well as give concrete decisions and next steps.
+
+### Scope:
+
+This ADR applies to, and should cover, all repositories within the Express Projects three orgs (`jshttp`, `pillarjs`,
+`expressjs`). While everything stems from support for the primary `express` package, due to the coupling of many of our
+dependency and middleware packages, we must also address those packages as well. 
+
+### ADR Goals:
+
+- Improve documentation on our support goals
+- Document criteria and reasoning for when to release new majors
+- Define life cycle for releases, terminology we will use, 
+- Improve documentation on how our LTS relates to Node.js version support
+- Documentation on how our LTS relates to our direct dependencies and other Express project repos/packages
+- Set clear next steps for documenting and publicising the LTS strategy 
+- Select dates for known release line support and EOL
+- Create starter language and graphics for our future blog posts, website additions, and other necessary docs 
+
+### ADR Non-Goals:
+
+- Create an end-user documentation resource (either for immediate use or future use)
+- Commit to future release dates
+
+**Existing issues/discussions/pull requests related to this:**
+
+- Initial LTS strategy doc: https://github.com/expressjs/discussions/blob/master/docs/LTS-strategy.md
+- Additional LTS doc proposal: https://github.com/expressjs/admin/pull/3
+- https://github.com/ctcpip/express-release
+- Some other related discussions:
+  - https://github.com/expressjs/discussions/issues/196
+  - https://github.com/expressjs/discussions/issues/172
+  - https://github.com/expressjs/discussions/issues/67
+- TC Slack Discussion: https://openjs-foundation.slack.com/archives/C06KMCETHG9/p1724863852681889
+
+### Package/Repository Types
+
+We maintain four primary types of packages/repos:
+
+1. `express`: The minimal http framework
+2. `express` dependencies: Libraries which are used directly or transitively within `express` and also by others outside
+   of the project
+3. Middleware: Libraries for use *with* `express` but not installed with it directly
+4. Others: Tools, documentation, websites, resources, etc.
+
+While much of this proposal is targeted specifically at `express`, this terminology will be used to differentiate where
+we are not only talking about `express`.
+
+## Decision
+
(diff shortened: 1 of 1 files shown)


docs/adr/lts-strategy.md
```diff
@@ -0,0 +1,176 @@
...
+  `express`, dependencies, middleware, & tools/other.
+- `express` version 4.x is considered a special case. It will receive a longer `MAINTENENCE` phase of 18 months or until
+  the TC agrees we are save cutting support.
+- `express` dependency packages will follow the same timeline and support dates for the `express` version which they are
+  included with.
+- Middleware packages should follow the timeline and support dates for the `express` versions they are compatible, but
+  should make reasonable and informed decisions which can deviate if necessary.
+- All other packages are suggested to define an appropriate strategy for support.
+
+### Node.js Version Support
```
wesleytodd, line 96: It is questionable if this section belongs in this ADR. IMO they are separate but have some connections. We do not want our support/release plan to conflict too much with Node.js so that we avoid supporting too many EOL Node.js versions. But other than that, maybe we just remove this section for now and make a separate "node.js version support strategy"?

docs/adr/lts-strategy.md
```diff
@@ -0,0 +1,176 @@
...
+
+This includes this doc, but also detailed things about the process that are not necessary for anyone but us. This
+documentation will live in one of three places:
+
+1. The `discussions` repo. If the content applies to more than one repo/package in any of the three orgs, is general in
+   nature, or otherwise has no logical other place to go it goes here.
+2. The individual repo/package to which it applies. If an individual package has more detailed or different needs the
+   docs will live along with the repo as decided by the captains.
+
+**User documentation:**
```
bjohansebas, line 161: 3?

```suggestion
3. **User documentation:**
```

wesleytodd: No, this is intended to be a new section like line 151 is. I debated using a fourth level of heading but this seemed to look alright in the web view.

docs/adr/lts-strategy.md
```diff
@@ -0,0 +1,176 @@
...
+requirements:
+
+- `express` major versions will go through three supported phases: Current, Active, & Maintenance
+  - `CURRENT`: A new major version is designated as `CURRENT` upon release. It is available but not the `latest` version
+    on npm for a minimum of 3 months.
+  - `ACTIVE`: After the minimum 3 month period and the TC agrees it is stable and secure, the `ACTIVE` version is
+    tagged `latest` on npm for a minimum of 12 months.
+  - `MAINTENANCE`: When a new major version becomes `ACTIVE`, the previous major version enters `MAINTENANCE` for 12 months.
+- After the `MAINTENANCE` time has ended the major version is considered `EOL` and is unsupported.
+- During the 12 month `ACTIVE` period, a new major version may be released but *not* until at least 6 months have passed
```
naugtur, line 80: Earlier the period is defined as "at least 12". Not sure if the intention here is to mirror it or say exact 12

docs/adr/lts-strategy.md
```diff
@@ -0,0 +1,176 @@
...
+  since the `CURRENT` version became `ACTIVE`.
+- Users are required to follow the head/latest of each major release line for support with all packages (`express`,
+  dependencies, middleware, & tools/other)
+- If we have no necessary breaking changes, we will not release a new major version. Applies to `express`,
+  dependencies, middleware, & tools/other.
+- Dropping old Node.js versions *alone* is not enough of a breaking change to release a new major version. Applies to
+  `express`, dependencies, middleware, & tools/other.
+- `express` version 4.x is considered a special case. It will receive a longer `MAINTENENCE` phase of 18 months or until
+  the TC agrees we are save cutting support.
```
naugtur, line 89:
```suggestion
  the TC agrees we are safe cutting support.
```

docs/adr/lts-strategy.md
naugtur: Public docs however would benefit of having both explained in one place.

docs/adr/lts-strategy.md
```diff
@@ -0,0 +1,176 @@
...
+  should make reasonable and informed decisions which can deviate if necessary.
+- All other packages are suggested to define an appropriate strategy for support.
+
+### Node.js Version Support
+
+- Each major version of all packages (`express`, deps, middleware, etc) will support a minimum of the oldest Node.js LTS
+at the time of release and maintain that support through its entire life cycle. This includes EOL Node.js versions.
+- `express` may choose to continue support for EOL Node.js versions in all three phases if deemed valuable, but will
+strive to cut support before it becomes a drain on progress.
+- All `express` dependency packages must support, at a minimum, the Node.js versions that are supported by the `express`
```
naugtur, line 102: The combination of this and preceding line might mean that when an unsupported version of node.js has a particular vulnerability, dependency packages are forced to provide workarounds to avoid exposing the vulnerability.

docs/adr/lts-strategy.md
```diff
@@ -0,0 +1,176 @@
...
+- Middleware packages must at a minimum follow the same Node.js versions as the `CURRENT` and `ACTIVE` release lines of
+`express`. They should attempt to also support `MAINTENANCE` lines as well, but the discretion of the captains is
+important.
+- All other packages are suggested to follow a reasonable Node.js support policy for their needs.
+
+### Life Cycle Phases & Meaning
+
+**CURRENT**
+
+- New majors will go through a short period hardening to ensure stability, security, and ecosystem libraries/resources
```
naugtur, line 113:
```suggestion
- New majors will go through a short period of hardening to ensure stability, security, and ecosystem libraries/resources
```

docs/adr/lts-strategy.md
naugtur: Nitpicky: Sighted folks use H levels for understanding document structure too. I'd go with H4 despite it's unfortunate rendering on GitHub.

bjohansebas: It seems odd to say that it will live in three places, but the list only includes two places.

docs/adr/lts-strategy.md
```diff
@@ -0,0 +1,176 @@
...
+**`express` dependencies**
+
+### Schedule
+
+For the existing release lines, we will set the following phase dates:
+
+
+| Major | CURRENT | ACTIVE | MAINTENANCE | EOL |
+| ----- | ------- | ------ | ----------- | --- |
+| 4.x   |         |        | 2025-04-01 | 2026-10-01 or later |
```
marco-ippolito, line 143: When v5 becomes tagged as latest, users running `npm i express` need to be aware they are using v5 and must know that should not use v4 which will be EOL after a certain date. The reason is that it will invalidate a lot of tutorials and beginner content, so it must be clear to users that they are installing v5 and v4 will not be supported.
Id start preparing the ecosystem for the end of v4 by adding deprecation warnings, blogposts etc...

docs/adr/lts-strategy.md
wesleytodd: Ah, good point and yes I agree fully. I will add that to the section about public docs.

docs/adr/lts-strategy.md
wesleytodd: Oh! ha, maybe I had three in mind and cut one. Yeah I will fix that.

docs/adr/lts-strategy.md
wesleytodd: Most of this was copied from conversations we had, so it was not all written at once and I probably should have re-read some after it was together. For this one, I actually think maybe we should just drop the months here so we dont have inconsistent language but also make clear it can happen during that period:

```suggestion
- During the `ACTIVE` period, a new major version may be released but *not* until at least 6 months have passed
```

docs/adr/lts-strategy.md
wesleytodd: Hm, interesting, yeah I think I follow but I also think that we have discussed something similar to this as well before. Our discussion was around v4 users stuck on EOL node versions and how our pass through apis expose many of the same vulns as the underling runtime. IIRC we decided this was out of scope for what we consider this projects problem to solve. 

Now that you are saying that though, I wonder if we should add this to our threat model doc? https://github.com/expressjs/security-wg/blob/main/docs/ThreatModel.md cc @expressjs/security-wg

docs/adr/lts-strategy.md
wesleytodd: I am going to write the blog post after following up on these reviews. I am not sure we will even deprecate on npm after that. That said, I agree for beginners this is not enough and we might want to do more. I agree this is going to be a *big* problem and it is one why we didn't go latest earlier. Do you think we should do a `postinstall` message or something? This is for sure a tough one to solve, so open to any recommendations.

marco-ippolito: yeah I think we should add a warning (not sure you can do it through postinstall since you cant write to stdout for sure) saying hey express 4 will be EOL this date, please migrate

ljharb: You could certainly `npm deprecate` every v4 version except the latest one, though.

docs/adr/lts-strategy.md
naugtur: I think it's ok to go the other way and explicitly say that express and it's ecosystem will not be forced to work around vulns in EOL Node.

docs/adr/lts-strategy.md
UlisesGascon, line 139:
```suggestion
| 4.x   |         |        | 2025-04-01 | 2026-10-01 |
```

We can always "negociate" but let's avoid setting expectation to delay that day forever :)

wesleytodd: We had discussed that at one point IIRC and I think we were unsure of the value, but since we won't be doing that for sure until EOL, we have 18 months to decide.

wesleytodd: This was at @ctcpip's request: https://openjs-foundation.slack.com/archives/C02QB1731FH/p1743197622883289?thread_ts=1743196832.158829&cid=C02QB1731FH

ljharb: The value would be to aggressively push people, at the least, to be on the latest v4, and at best, v5.

bjohansebas (approved):

jonchurch:
I had dropped this draft of messaging around EOL into the WG 
https://github.com/expressjs/discussions/issues/351#issuecomment-2767320770

docs/adr/lts-strategy.md
ctcpip: We briefly discussed today, and we have agreed to remove some of this language from this doc, to keep it focused specifically on release strategy and to remove excess details that do not need to be explicitly stated or promised.

docs/adr/lts-strategy.md
```diff
@@ -0,0 +1,173 @@
...
+
+- New majors will go through a short period of hardening to ensure stability, security, and ecosystem libraries/resources
+  compatibility.
+- We will strive to ensure no breaking changes are included, but reserve the right to make security or high priority
+  fixes of breaking nature within this period.
+- `CURRENT` lines will receive all types of active work including: bug fixes, security patches, new features, and
+  deprecation notices.
+- Users are recommended to use  `CURRENT` lines and to upgrade as quickly as their risk profile allows
+
+#### ACTIVE
```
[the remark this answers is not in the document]
ctcpip: I don't think we benefit from specifying anything particular here, because if we ever had a potential non-semver breaking change, it would be evaluated very carefully, and only done if the benefit truly outweighed the harm it would cause.

docs/adr/lts-strategy.md
ctcpip: We discussed briefly today, but this is somewhat out of scope for what we are trying to do with this document.  Let's create a separate issue to track strategy around v4 EOL planning.

ctcpip: If we are setting a date, we should have reasonable confidence that we would actually meet that date.  Otherwise, it's just completely arbitrary.  I suggest we leave as is, or put `TBD`.

Phillip9587 (approved):

docs/adr/lts-strategy.md
wesleytodd: I removed this section and called out our plan to address this independently.

docs/adr/lts-strategy.md
wesleytodd: Updated this section to add more clarity, we can continue the discussion around v4 EOL plans as @ctcpip said in a follow up issue.

wesleytodd: Changed to TBD with a bit of extra language in the footnote.

docs/adr/lts-strategy.md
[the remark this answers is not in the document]
wesleytodd: Yeah, I am alright leaving this unaddressed for now. Can we call this resolved and treat any additions as future edits after merging this?

UlisesGascon (approved):

docs/adr/lts-strategy.md
```diff
@@ -0,0 +1,172 @@
...
+| ----- | ------- | ------ | ----------- | --- |
+| 4.x   |         |        | 2025-04-01 | *no sooner than 2026-10-01 |
+| 5.x   | 2024-09-11 | 2025-03-31 | **TBD | **TBD |
+| 6.x   | ***TBD | | | |
+
+*: v4 is a special case, and we may extend MAINTENENCE support. This date is called out to give users confidence we will
+   not end-of-life *earlier* than this date.
+**: v5 MAINTENENCE and EOL dates are determined by when v6 is released. We will update them when v6 release dates are
+    committed to. 
+***: v6 work has not started yet. The earliers we could choose to release by this proposal is 2026-01-01. We will update
```
bjohansebas, line 139: Just a small detail, I think you need to escape each * (\*) so that markdown doesn’t apply its formatting, that happened on the website, and it will very likely happen here too.

bjohansebas (approved):

docs/adr/lts-strategy.md
wesleytodd: Screenshot 2025-04-17 at 3 12 56 PM

Yeah, looks like the formatting is broken. I bet once I move it to a list the stars will be an issue. I will make them numbered footnotes.

wesleytodd: Screenshot 2025-04-17 at 3 16 40 PM

Fixed.

docs/adr/lts-strategy.md
```diff
@@ -0,0 +1,172 @@
...
+
+The Express project will have a well defined Long Term Support (LTS) strategy and schedule based on the following
+requirements:
+
+- `express` major versions will go through three supported phases: Current, Active, & Maintenance
+  - `CURRENT`: A new major version is designated as `CURRENT` upon release. It is available but not the `latest` version
+    on npm for a minimum of 3 months.
+  - `ACTIVE`: After the minimum 3 month period and the TC agrees it is stable and secure, the `ACTIVE` version is
+    tagged `latest` on npm for a minimum of 12 months.
+  - `MAINTENANCE`: When a new major version becomes `ACTIVE`, the previous major version enters `MAINTENANCE` for 12 months.
```
ctcpip, line 78:
```suggestion
  - `MAINTENANCE`: When a new major version becomes `ACTIVE`, the previous major version enters `MAINTENANCE` for a minimum of 12 months.
```

ctcpip (approved):

PR #367, opened 2025-04-30
fix: ADR cleanup

wesleytodd:

### DIFF
--- a/docs/adr/000-adr-template.md
+++ b/docs/adr/000-adr-template.md
@@ -4,7 +4,7 @@ This is the base template that we use
 
 ```md
 
-# ADR [Number]: [Title of Decision]
+# ADR [PR Number]: [Title of Decision]
 
 ## Status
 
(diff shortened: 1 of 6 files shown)


ctcpip:
hmmm... hate to say it but I think we need an ADR for numbering...  UUID?  monotonically increasing integer starting from 1?  PR number?  🤔🤔🤔🤔🤔🤔🤔🤔🤔

wesleytodd:
I was just kidding because of our exchange, but I really do think PR number is a reasonable way. No need to go look it up when merging (and forget, like I did). At least these numbers mean something.

bjohansebas (approved):

UlisesGascon (approved):

ctcpip:
slight dispreference due to not knowing in advance what the PR number will be. (or guessing and risking race condition), meaning a follow-on commit will always be required

also not convinced we need numeric/symbolic identifiers at all, but I have little appetite to bikeshed this

wesleytodd:
> also not convinced we need numeric/symbolic identifiers at all, but I have little appetite to bikeshed this

Ok, you have me sold on this. I almost did this in the PR but realized that it meant the file names might become more important.

Honestly I just did this as a joke, we can even close it if we just want to let the lord of chaos reign.

ctcpip (approved):

PR #375, opened 2025-05-15
clarify nomination process for Technical Committee members

bjohansebas:
Just a small rewording to make it clear that only TC members can nominate others to become TC members

> [!IMPORTANT]
> If this is approved by the TC, an issue must be opened in the CPC for their approval as well, since it is a change to the CHARTER


UlisesGascon (approved):

ctcpip (changes requested):
per TC discussion, blocking for now as we work through how best to deal with this

wesleytodd (changes requested):
As discussed:

1. We should handle these nominations in a consistent way across all roles
2. Nominations should be done in private, suggest reaching out in slack or email

We may need to follow up with other docs to apply this process in the right places. Those don't need to block this PR, but we need to at least reflect the private nomination in this PR.

bjohansebas:
Also, the other reason this PR was being blocked was because we needed to decide whether to include it in the charter or in the governance file.

docs/CHARTER.md
```diff
@@ -51,8 +51,8 @@ Section Intentionally Left Blank
 ## Section 3: Express Governing Body
 
 The Express project is managed by the Technical Committee ("TC").
-Members can be added to the TC at any time. Any committer can nominate
-another committer to the TC and the TC uses its standard consensus
+Members can be added to the TC at any time. Any committer can privately nominate
+another committer to the TC, preferably slack or alternatively by email, and the TC uses its standard consensus
```
jonchurch:
```suggestion
Members can be added to the TC at any time. Any committer can nominate
another committer to the TC, by privately contacting one or more TC members, and the TC uses its standard consensus
```

UlisesGascon: There is a consensus on the TC meeting for this changes. So I will commit it.

jonchurch:
CPC has agreed (per @ctcpip) that this process piece does indeed belong in the `Charter.md`

To get CPC approval, wait to merge, batch any pending Charter changes that we may have outstanding (there's at least one other Charter change pending I believe), then get the Diff over to the CPC which will take _time_ to review and approve by that body. 


We can repurpose this PR to be a Charter change PR in general if required. cc @bjohansebas

wesleytodd:
Changed target branch to `governance-updates` so that we can land these and then take *one* consolidated branch to the CPC for approval.

jonchurch (approved):

jonchurch:
@wesleytodd private nomination is reflected in this PR, and @ctcpip we are landing these governance changes to a `governance-update` branch per wes' target change

Are yall's request changes review satisfied?

Im tracking the batched changes getting sent to CPC in #466 which is now on the agenda instead of the individual ones we have discussed enough times

wesleytodd (approved):

PR #387, opened 2025-05-29
docs: clarify release authorization and permissions in release process

bjohansebas:
If two captains are not part of the TC, and the other captain gives authorization to release the package, is that valid? Or does there always have to be authorization from the TC?

### DIFF
--- a/docs/contributing/release-process.md
+++ b/docs/contributing/release-process.md
@@ -7,7 +7,10 @@ package governed by Express.
 
 ## Who can make releases?
 
-According to project governance, only package Captains or members of the Technical Committee (TC) are permitted to make releases. Captains have the freedom to make releases for their own package whenever necessary, while TC members may make releases for any package as needed.
+According to project governance, only package Captains or members of the Technical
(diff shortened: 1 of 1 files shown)


bjohansebas:
cc: @expressjs/express-tc

docs/contributing/release-process.md
```diff
@@ -2,25 +2,22 @@
...
+According to project governance, only package Captains or members of the Technical Committee (TC) are permitted to make releases. Captains have the freedom to make releases for their own package whenever necessary, while TC members may make releases for any package as needed.
 
 ### 1. Github release access
 
-The individual making the release will need to be a member of the
-expressjs/express team with Write permission level so they are able to tag the
-release commit and push changes to the expressjs/express repository
-(see Steps 4 and 5).
+The person making the release must be either the package captain or a member of the TC in
+order to have sufficient permissions to do the work.
```
bjohansebas, line 15: This change is because it was referring only to Express and not in general

wesleytodd (approved):

PR #389, opened 2025-05-30
remove triage meeting

bjohansebas:
The triage meeting is removed, as there is no attendance and there aren't really any topics that require a meeting, they can be resolved asynchronously.

cc: @expressjs/triagers


bensternthal:
@UlisesGascon I removed this event from LFX and from the public calendar.

UlisesGascon (approved):

PR #391, opened 2025-06-03
docs: add ADR for branch protection

UlisesGascon:
Related https://github.com/expressjs/security-wg/issues/2

cc: @expressjs/express-tc @expressjs/security-wg

### DIFF
--- a/docs/adr/367-branch-protection-md
+++ b/docs/adr/367-branch-protection-md
@@ -14,7 +14,7 @@ Accepted
 
 ## Context
 
-Branch protection rules in GitHub enforce key policies—such as required reviews, status checks, and linear history—that help safeguard critical branches like `main`, `master` or version specific. Despite the clear benefits, adoption across repositories has been inconsistent, leading to:
+Branch protection rules in GitHub enforce key policies—such as required reviews, status checks, and linear history—that help safeguard critical branches like `main`, `master` or version branches/tags. Despite the clear benefits, adoption across repositories has been inconsistent, leading to:
 
 - Risk of direct or accidental pushes to protected branches
 - Increased likelihood of unreviewed or unsafe code merges
--- a/docs/adr/367-branch-protection-md
+++ b/docs/adr/367-branch-protection-md
@@ -36,7 +36,7 @@ Key rules to be applied:
 
 Optional rules:
 - Enforce linear history (disable merge commits, allow squash/rebase)
-- Restrict who can push directly to the protected branches
+- Restrict who can push directly to the protected branches to people who can release (captains, TC, etc)
 
 ### Exceptions:
 
--- a/docs/adr/367-branch-protection-md
+++ b/docs/adr/367-branch-protection-md
@@ -30,7 +30,7 @@ This decision aims to normalize a secure-by-default stance while maintaining fle
 
 Key rules to be applied:
 
-- Protect the `main`, `master`, and `v*` branches
+- Protect release branches, this includes: `main`, `master`, and `v*` branches/tags
 - Require pull request reviews before merging (at least 1 approval)
 - Require all status checks to pass before merging
 
--- a/docs/adr/367-branch-protection-md
+++ b/docs/adr/367-branch-protection-md
@@ -43,6 +43,9 @@ Optional rules:
 - TC members and Repo Captains will retain the ability to **temporarily bypass branch protections** in critical scenarios:
   - Security fixes that require immediate deployment
   - Emergency production incidents
+  - When CI is clearly broken for unrelated reasons
+  - To fix history when necessary
+  - When a reasonable case is made where the exception makes sense
 - These exceptions must be documented in the relevant PR or release communication
 
 ## Rationale
--- a/docs/adr/367-branch-protection-md
+++ b/docs/adr/367-branch-protection-md
@@ -1,0 +1,106 @@
+# ADR 367: Enforce Branch Protection by Default Across All Repositories
+
+## Status
+
+Accepted
+
+## Submitters
+
+- @UlisesGascon
+
+## Decision Owners
+
+- @expressjs/express-tc
+
+## Context
+
+Branch protection rules in GitHub enforce key policies—such as required reviews, status checks, and linear history—that help safeguard critical branches like `main`, `master` or version specific. Despite the clear benefits, adoption across repositories has been inconsistent, leading to:
+
+- Risk of direct or accidental pushes to protected branches
+- Increased likelihood of unreviewed or unsafe code merges
+- Reduced ability to enforce standardized development workflows
+
+GitHub allows fine-grained control over branch protections, including the ability for administrators (tc members, repo captains...) to bypass rules when needed (e.g., urgent hotfixes or security patches).
+
+This decision aims to normalize a secure-by-default stance while maintaining flexibility for urgent, responsible exceptions.
+
+## Decision
+
+**Branch protection rules will be enabled by default on all repositories.**
+
+Key rules to be applied:
+
+- Protect the `main`, `master`, and `v*` branches
+- Require pull request reviews before merging (at least 1 approval)
+- Require all status checks to pass before merging
+
+Optional rules:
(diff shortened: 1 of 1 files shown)


docs/adr/367-branch-protection-md
```diff
@@ -0,0 +1,106 @@
...
+- Possible misuse of admin bypass if not documented
+
+
+## Implementation
+
+### Phase 1: Rollout to Active Repos
+
+- Identify all active repositories
+- Apply default protection rules using GitHub API or organization policies
+- Notify repo maintainers of the change
```
ctcpip, line 92:
```suggestion
- Identify all active repositories
- Notify repo maintainers of the change
- Apply default protection rules using GitHub API or organization policies
```

docs/adr/367-branch-protection-md
```diff
@@ -0,0 +1,106 @@
...
+
+This decision aims to normalize a secure-by-default stance while maintaining flexibility for urgent, responsible exceptions.
+
+## Decision
+
+**Branch protection rules will be enabled by default on all repositories.**
+
+Key rules to be applied:
+
+- Protect the `main`, `master`, and `v*` branches
```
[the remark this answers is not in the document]
wesleytodd:
```suggestion
- Protect release branches, this includes: `main`, `master`, and `v*` branches/tags
```

This way we are inclusive of things that may have other patterns like the website without having to be exhaustive.

docs/adr/367-branch-protection-md
```diff
@@ -0,0 +1,106 @@
...
+- Require pull request reviews before merging (at least 1 approval)
+- Require all status checks to pass before merging
+
+Optional rules:
+- Enforce linear history (disable merge commits, allow squash/rebase)
```
wesleytodd, line 38: Big fan of this, and I think we also had earlier discussion about this where we all generally agreed on it.

docs/adr/367-branch-protection-md
```diff
@@ -0,0 +1,106 @@
...
+- Restrict who can push directly to the protected branches
+
+### Exceptions:
+
+- Tc members and repo captains will retain the ability to **temporarily bypass branch protections** in critical scenarios:
+  - Security fixes that require immediate deployment
+  - Emergency production incidents
```
wesleytodd, line 45:
```suggestion
  - Emergency production incidents
  - When CI is clearly broken for unrelated reasons
  - To fix history when necessary
  - When a reasonable case is made where the exception makes sense
```

docs/adr/367-branch-protection-md
```diff
@@ -0,0 +1,106 @@
...
+### Phase 1: Rollout to Active Repos
+
+- Identify all active repositories
+- Apply default protection rules using GitHub API or organization policies
+- Notify repo maintainers of the change
+
+### Phase 2: Harden Governance
+
+- Periodic review of admin bypass usage using the Scorecard Monitor ([ref](https://github.com/expressjs/security-wg/pull/70))
+- Enforce admin bypass documentation for emergency changes
```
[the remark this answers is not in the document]
wesleytodd: I added a comment about adding more valid reasons above, but seeing this comment I think I agree that we should make that an amendment onto this larger ADR. For a while our CI was broken and the only way I could merge things was bypassing the protections. I think it is unreasonable to limit ourselves like this until we have automated releases since we can always fix the history if a mistake is made.

wesleytodd (changes requested):
Most of my comments are nit picks, but I do agree with @ctcpip that we should spin out the bypass stuff into a follow up. It can even be an edit to this after we land IMO, but I like the main content of this so it would be unfortunate to hold that up because we need to discuss one small part.

docs/adr/367-branch-protection-md
UlisesGascon: True, We can do it in a separate ADR?

docs/adr/367-branch-protection-md
```diff
@@ -0,0 +1,109 @@
...
+- Require pull request reviews before merging (at least 1 approval)
+- Require all status checks to pass before merging
+
+Optional rules:
+- Enforce linear history (disable merge commits, allow squash/rebase)
+- Restrict who can push directly to the protected branches to people who can release (captains, TC, etc)
+
+### Exceptions:
+
+- TC members and Repo Captains will retain the ability to **temporarily bypass branch protections** in critical scenarios:
```
bjohansebas, line 43: And how will this be done? Remember that the captains only have maintainer permissions, which means they can't edit branch protections (only admins can do that). I think there's an option to allow certain users to bypass protections, but I'm not sure if it can be done by teams, which would be the best approach.

ctcpip: it would be done by someone with admin privileges to temporarily enable bypassing

docs/adr/367-branch-protection-md
```diff
@@ -0,0 +1,109 @@
...
+  - Security fixes that require immediate deployment
+  - Emergency production incidents
+  - When CI is clearly broken for unrelated reasons
+  - To fix history when necessary
+  - When a reasonable case is made where the exception makes sense
```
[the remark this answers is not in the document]
UlisesGascon: I see that issue quite often in other repos too, but the solution is not bypass 😅

docs/adr/367-branch-protection-md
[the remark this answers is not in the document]
jonchurch: Chris was speaking specifically about keeping the Branch Protection ADR as its own, and a score card ADR as a separate decision.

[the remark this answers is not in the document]
wesleytodd: I think all the other changes were made, and this one only says "enforce" the rules we document above. I think that means we can mark it resolve. I am going to and change to approve, but as it was @ctcpip's comment I am very much alright with re-opening it to continue discussion if necessary.

wesleytodd (approved):

docs/adr/367-branch-protection-md
ljharb: unfortunately tho there's issues with this; linear history is good, but github's rebasemerge/squashmerge is bad.

bjohansebas (approved):

UlisesGascon:
So... I plan to land this next week. Please add your reviews or blockers :)

docs/adr/367-branch-protection-md
```diff
@@ -0,0 +1,109 @@
...
+
+Key rules to be applied:
+
+- Protect release branches, this includes: `main`, `master`, and `v*` branches/tags
+- Require pull request reviews before merging (at least 1 approval)
+- Require all status checks to pass before merging
+
+Optional rules:
+- Enforce linear history (disable merge commits, allow squash/rebase)
+- Restrict who can push directly to the protected branches to people who can release (captains, TC, etc)
```
blakeembrey, line 39: Can you help me understand this optional rule and how it interacts with the exception below?

From my understanding enabling this would allow the people who can release to permanently bypass the branch protection rules, rather than temporarily. It'd be simpler to have everything follow the same process to avoid any confusion when it comes down to this.

docs/adr/367-branch-protection-md
```diff
@@ -0,0 +1,109 @@
...
+- Possible misuse of admin bypass if not documented
+
+
+## Implementation
+
+### Phase 1: Rollout to Active Repos
+
+- Identify all active repositories
+- Notify repo maintainers of the change
+- Apply default protection rules using GitHub API or organization policies
```
blakeembrey, line 95: CleanShot 2025-07-13 at 11 20 40@2x

Can you document the API you use so people can re-use/update it over time? 🙏

blakeembrey (approved):
Thanks for creating this and following up for comments, I missed the full discussion earlier.

docs/adr/367-branch-protection-md
```diff
@@ -0,0 +1,109 @@
+# ADR 367: Enforce Branch Protection by Default Across All Repositories
```
blakeembrey, line 1: Should this be `ADR 391`?

PR #398, opened 2025-06-19
meetings: remove PR and push triggers, add workflow_dispatch

jonchurch:
Idk if we need/want to run it on PR and push?

At the least though, adding the workflow_dispatch option would be very useful so we can run them on demand

### DIFF
--- a/.github/workflows/meetings.yml
+++ b/.github/workflows/meetings.yml
@@ -1,9 +1,6 @@
 name: Schedule Meetings
 on:
-  pull_request:
-  push:
-    branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/meetings.yml
```diff
@@ -1,9 +1,6 @@
 name: Schedule Meetings
 on:
-  pull_request:
```
bjohansebas, line 3: I think a pull request makes sense when changes are made to the workflow.

bjohansebas (approved):

PR #408, opened 2025-07-21
fix: TC to meet 2 hours earlier

wesleytodd:
We have been discussing moving the meetings to an earlier time. This proposes 2 hours earlier.

PR #413, opened 2025-08-04
Fix: TC Schedule time

UlisesGascon:
Fix #408 as the calendar only reflected 1h and not 2h (discussed on Slack).

### DIFF
--- a/.github/workflows/meetings.yml
+++ b/.github/workflows/meetings.yml
@@ -25,8 +25,8 @@ jobs:
           agendaLabel: 'tc agenda'
           meetingLabels: 'meeting'
           # https://github.com/expressjs/discussions/issues/195#issuecomment-1973732769
-          # Starting on 2024-03-18 at 9pm UTC (2024-03-18T21:00:00.0Z) with a period of 2 weeks (P2W)
-          schedules: '2024-03-18T19:00:00.0Z/P2W'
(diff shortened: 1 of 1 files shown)


ctcpip (changes requested):
wait, this will actually make the issues the wrong time again!  🤣 

let's discuss at mtg

.github/workflows/meetings.yml
```diff
@@ -25,8 +25,8 @@ jobs:
           agendaLabel: 'tc agenda'
           meetingLabels: 'meeting'
           # https://github.com/expressjs/discussions/issues/195#issuecomment-1973732769
-          # Starting on 2024-03-18 at 9pm UTC (2024-03-18T21:00:00.0Z) with a period of 2 weeks (P2W)
-          schedules: '2024-03-18T19:00:00.0Z/P2W'
+          # Starting on 2024-03-18 at 6pm UTC (2024-03-18T18:00:00.0Z) with a period of 2 weeks (P2W)
+          schedules: '2024-03-18T18:00:00.0Z/P2W'
```
blakeembrey, line 29: Let's add a `timeZone` option instead to the library? Otherwise it'll need changing again in November...

blakeembrey: Opened https://github.com/pkgjs/meet/pull/225.

PR #424, opened 2025-08-21
docs: documentation for third-party services used in the project

### LINKED ISSUE #324: Proposal: Create an infrastructure file
It would be very useful to document in detail which resources Express is using and who is providing them. This would not only ensure greater transparency in resource management but also be extremely helpful for future maintainers of the project. Having a clear record would allow them to easily understand the current infrastructure.

For example:

- https://github.com/curl/curl/blob/master/docs/INFRASTRUCTURE.md

bjohansebas:
Added a comprehensive list of third-party services used by the project, that’s the information I have; I’m not sure if we use any other services as part of our infrastructure, so please let me know closes https://github.com/expressjs/discussions/issues/324

cc: @bensternthal @expressjs/express-tc

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -9,7 +9,7 @@ This project relies on the following third-party services:
 - **DNS**: LFX
   - **Domain Names**: expressjs.com
 - **Password Management:** 1Password
-- **Funding Management:** 
+- **Funding Management:**
   - [Open Collective](https://opencollective.com/express)
   - Thanks.dev
 - **Translations Management:** Crowdin
@@ -20,7 +20,7 @@ This project relies on the following third-party services:
 - **Code Repository:** GitHub (expressjs, pillarjs, jshttp)
 - **CI/CD:** GitHub Actions
 - **Website Preview:** Netlify
-- **Artifact Storage:** 
+- **Artifact Storage:**
   - Docker Hub
   - NPM (expressjs, express, jshttp, pillarjs)
 
--- a/docs/services.md
+++ b/docs/services.md
@@ -1,0 +1,38 @@
+# Third-Party Services Used
+
+This project relies on the following third-party services:
+
+## Hosting & Infrastructure
+
+- **Hosting**: Github Pages, Cloudflare
+- **Analytics**: Cloudflare
+- **DNS**: LFX
+  - **Domain Names**: expressjs.com
+- **Password Management:** 1Password
+- **Funding Management:** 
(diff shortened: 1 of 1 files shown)


docs/services.md
```diff
@@ -0,0 +1,38 @@
...
+## Hosting & Infrastructure
+
+- **Hosting**: Github Pages, Cloudflare
+- **Analytics**: Cloudflare
+- **DNS**: LFX
+  - **Domain Names**: expressjs.com
+- **Password Management:** 1Password
+- **Funding Management:** 
+  - [Open Collective](https://opencollective.com/express)
+  - Thanks.dev
```
bjohansebas, line 14: Thanks.dev should be connected to the Open Collective we have, because if the funding isn’t collected it’s lost, as thanks.dev refunds it to the sponsor after three months

docs/services.md
```diff
@@ -0,0 +1,38 @@
...
+- **Translations Management:** Crowdin
+- **Cloud Providers:** Digital Ocean, Datadog
+
+## Development & CI/CD
+
+- **Code Repository:** GitHub (expressjs, pillarjs, jshttp)
+- **CI/CD:** GitHub Actions
+- **Website Preview:** Netlify
+- **Artifact Storage:** 
+  - Docker Hub
```
bjohansebas, line 24: As far as I remember, @UlisesGascon created a Docker account for the perf team

docs/services.md
```diff
@@ -0,0 +1,38 @@
...
+This project relies on the following third-party services:
+
+## Hosting & Infrastructure
+
+- **Hosting**: Github Pages, Cloudflare
+- **Analytics**: Cloudflare
+- **DNS**: LFX
+  - **Domain Names**: expressjs.com
+- **Password Management:** 1Password
+- **Funding Management:**
```
bjohansebas, line 12: I know some packages are under the Tidelift program. I’m not sure if we should add it.

jonchurch:
We are discussing making a private repo to house this kind of information, in order to help with bus factor stuff or hand off. We can share it with foundation folks.

AKA, lets create a list of our various accounts and maybe the email addresses that own them (although 1Password theoretically should serve as a self documenting list of accounts we own and the email addresses associated with them if it works that way)

wesleytodd:
Not going to re-open this, but we should try to do this against the new `.github-private` repo: https://github.com/expressjs/.github-private

PR #429, opened 2025-08-27
🕰️ add timezones to meeting template

ctcpip:
I wanted to add UTC and figured it would be a good opportunity to also add some timezones to reflect meeting attendees.  (I realize the added European countries all share the same timezone. Don't `@` me!  😄)


sheplu (approved):
LGTM

PR #444, opened 2025-11-03
Default workflows part of all repositories

sheplu:
RFC to standardize the minimal set of workflows expected on all repositories

### DIFF
--- a/docs/rfc/0003-core-workflows.md
+++ b/docs/rfc/0003-core-workflows.md
@@ -1,0 +1,79 @@
+# Standard default CI/CD workflows for Express.js repositories
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


docs/rfc/0003-core-workflows.md
```diff
@@ -0,0 +1,79 @@
...
+## Prior Art
+
+- GitHub recommends reusable organization workflows for consistency and maintenance.
+
+## Unresolved Questions
+
+- Which Node.js versions must be tested by default? (e.g., 22 + 24 vs 18 + 20 + 22 + LTS)  
+- Should push validation run on all branches or exclude `main`?  
+- Naming convention for workflows (`ci.yml`, `test.yml`, `validate.yml`)?  
+- Should `npm audit` or dependency checks be mandatory or optional?
```
bjohansebas, line 79: It definitely should be optional. All our dependencies are our own, and adding npm audit will only create noise in the workflow

ljharb: It’s not noise if it alerts about unaddressed CVEs tho

sheplu:
@expressjs/express-tc to be reviewed, definitely not priority 1 but this can spark a few discussion about the minimal requirement for all repos

PR #455, opened 2025-12-03
Rfc for reusable workflows naming repository

sheplu:
Rfc for reusable workflows naming repository

### DIFF
--- a/docs/rfc/0002-workflows-governance.md
+++ b/docs/rfc/0002-workflows-governance.md
@@ -1,0 +1,157 @@
+# Naming and Governance for the Central Repository of Reusable GitHub Actions
+
+## Summary
+
+This RFC proposes selecting an official repository name for hosting all shared and reusable GitHub Actions workflows used across the Express.js organization.
+
+This RFC also proposes the creation of a new **Infrastructure Working Group (WG)** responsible for governing CI/CD standards, maintaining shared workflows, and ensuring consistent automation practices across all Express.js repositories.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


sheplu:
@expressjs/express-tc to be reviewed. in itself that not blocking and can be changed quickly - just some extra time later one with changing that value everywhere :)

docs/rfc/0002-workflows-governance.md
```diff
@@ -0,0 +1,157 @@
+# Naming and Governance for the Central Repository of Reusable GitHub Actions
+
+## Summary
+
+This RFC proposes selecting an official repository name for hosting all shared and reusable GitHub Actions workflows used across the Express.js organization.
+
+This RFC also proposes the creation of a new **Infrastructure Working Group (WG)** responsible for governing CI/CD standards, maintaining shared workflows, and ensuring consistent automation practices across all Express.js repositories.
```
blakeembrey, line 7: Should this be part of this RFC? I think I'd block on this, I don't think we have enough "work" for a working group, but happy to be wrong.

blakeembrey (changes requested):
Going over the existing repos, we already have:

* https://github.com/expressjs/.github
* https://github.com/expressjs/secure-publish
* https://github.com/expressjs/tooling
* https://github.com/expressjs/ci-workflows

I'd be in favor of using `.github`, and `tooling` if there's consensus on avoiding `.github` for some reason, and delete the other two.

Finally, what about using one repo per workflow to make versioning simpler? This doc doesn't really address how you'd version workflows if they're all in one repo.

PR #496, opened 2026-05-16
feat: propose sponsors level for opencollective and github

sheplu:
RFC to define the different level of sponsoring we want to have / provide in both OpenCollective but also Github Sponsor

### DIFF
--- a/docs/rfc/0001-funding-tiers.md
+++ b/docs/rfc/0001-funding-tiers.md
@@ -1,0 +1,247 @@
+# Funding Tiers and Sponsor Rewards
+
+## Summary
+
+Replace Express's current flat two-tier funding structure (Backer $5/mo, Sponsor $100/mo, both with no described rewards) with an eight-tier ladder — Backer, Supporter, Patron, Bronze, Silver, Gold, Platinum, Diamond — each carrying explicit, escalating recognition rewards (SPONSORS.md → README logo → expressjs.com → docs). Top tiers use limited slot counts to create scarcity, with a single-slot Diamond tier as the apex. The same tier structure is mirrored across the project's funding surfaces (OpenCollective and GitHub Sponsors) so sponsors can fund Express through whichever platform their organization prefers. This RFC complements (and does not modify) [ADR 299: Funding Use Guidelines](../adr/299-funding-guidelines.md), which governs how funds are spent; this RFC addresses how sponsors are recognized.
+
+## Motivation
+
+Express is the most-depended-on web framework in the Node.js ecosystem, with over **100 million weekly downloads** on npm. Despite that footprint, the project's funding surfaces are underdeveloped:
+
+- Only **two recurring tiers** exist: Backer at $5/mo and Sponsor at $100/mo.
+- Neither tier describes any reward, recognition, or benefit.
+- There is no public sponsor list, no logo placement, and no differentiation by contribution level.
+
+This creates two concrete problems:
+
+1. **Corporate sponsors cannot justify their support internally.** Marketing, DevRel, and developer-experience teams who would happily back Express need legible recognition (a logo, a placement, a public list) to expense it. Today, Express offers nothing to point to. Peer projects do — and they attract more sponsors as a result.
+2. **There is no path above the entry sponsor tier.** $100/mo is the entry sponsor price across nearly every comparable project. Express has no tier above it, so larger organizations who want to support the project at a higher level have nowhere to go.
+
+The expected outcome is a more predictable level of recurring sponsor support that helps fund what ADR 299 already authorizes — TC member living expenses, travel, and equipment — without requiring per-event fundraising.
+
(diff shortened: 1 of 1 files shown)


sheplu:
@expressjs/express-tc tag to be aware before the next tc meeting

docs/rfc/0001-funding-tiers.md
```diff
@@ -0,0 +1,247 @@
+# Funding Tiers and Sponsor Rewards
+
+## Summary
+
+Replace Express's current flat two-tier funding structure (Backer $5/mo, Sponsor $100/mo, both with no described rewards) with an eight-tier ladder — Backer, Supporter, Patron, Bronze, Silver, Gold, Platinum, Diamond — each carrying explicit, escalating recognition rewards (SPONSORS.md → README logo → expressjs.com → docs). Top tiers use limited slot counts to create scarcity, with a single-slot Diamond tier as the apex. The same tier structure is mirrored across the project's funding surfaces (OpenCollective and GitHub Sponsors) so sponsors can fund Express through whichever platform their organization prefers. This RFC complements (and does not modify) [ADR 299: Funding Use Guidelines](../adr/299-funding-guidelines.md), which governs how funds are spent; this RFC addresses how sponsors are recognized.
```
wesleytodd, line 5: > escalating recognition rewards (SPONSORS.md → README logo → expressjs.com → docs)

I think we discussed the logo/sponsors pages and explicitly did not want to do a "pay for play" scheme. We wanted sponsors who *just* want to support the project because it is valuable to them for it to be secure and stable.

I personally am indifferent on this topic, but I think there is merit in making sure we don't end up with a bunch of sponsors on there just to get the logo/link back. Maybe we can find a tier setup that doesn't include these types of "rewards"?

docs/rfc/0001-funding-tiers.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Motivation
+
+Express is the most-depended-on web framework in the Node.js ecosystem, with over **100 million weekly downloads** on npm. Despite that footprint, the project's funding surfaces are underdeveloped:
```
wesleytodd, line 9:
```suggestion
Express is the most depended on web framework in the Node.js ecosystem, with over **100 million weekly downloads** on npm. Despite that footprint, the project's funding is almost non-existent:
```

docs/rfc/0001-funding-tiers.md
```diff
@@ -0,0 +1,247 @@
...
+
+- Only **two recurring tiers** exist: Backer at $5/mo and Sponsor at $100/mo.
```
wesleytodd, line 11: I would add that there are not many people at either of these tiers. Especially considering if you compared this with other larger or important projects (see below). Considering this is the section justifying work being done on a funding model, I think you should make the argument on *why* does the project even consider this important before focusing on the small details like this. Maybe just swapping the order of these to put this below? Like:

1. we have no sponsors
2. compare to other projects
3. why we think this is the case

docs/rfc/0001-funding-tiers.md
```diff
@@ -0,0 +1,247 @@
...
+| **Patron** | $50 | Name + patron badge in `SPONSORS.md` + name on the expressjs.com sponsors page | unlimited |
+| **Bronze Sponsor** | $100 | Small logo in `README.md` of `expressjs/express` + name on the expressjs.com sponsors page | unlimited |
+| **Silver Sponsor** | $500 | Medium logo in `README.md` + logo on the expressjs.com sponsors page | unlimited |
+| **Gold Sponsor** | $1,500 | Logo on the **expressjs.com homepage** + larger logo in `README.md` | **5 slots** |
+| **Platinum Sponsor** | $5,000 | Logo on the expressjs.com homepage **and** docs site header + top placement in `README.md` | **3 slots** |
+| **Diamond Sponsor** | $10,000 | Exclusive top placement on `README.md`, expressjs.com homepage, and docs site header. Dedicated **case study card** on `expressjs.com/sponsors` (see Diamond case study card below). Recognition as Express's principal sponsor on each funding surface. | **1 slot** |
+
+A **Custom Donation** option (one-time or recurring, any amount) remains available for contributors who do not match a tier.
+
+### Reward specifics
```
wesleytodd, line 48: This is going to be the most contentious part of this proposal. Honestly I would consider making this section more of a brainstorm idea section than a concrete proposal. That way we could discuss and land a less contentious "we want to revamp the OC levels" agreement first, then add on the specifics of the "tier benefits" (a name I would prefer over awards) later.

wesleytodd (commented):
Generally in favor of a revamp here, I just don't think we should get so far in the weeds as this proposal goes. In fact I just stopped reading at line 48 because there was a lot of words and it seemed fairly repetitive after I raised my concern about the reward structure.

I would consider breaking this up, and just generally reducing the verbosity used to describe the tiers proposed. Maybe just a single table with the proposed levels and short list of "benefits"?

Issue #459, opened 2025-12-18
Use Node.js built-in debug and deprecated utilities instead of relying on external packages.


bjohansebas:
Since this is a change I would like to propose for major versions of the packages, I am opening this discussion here.

Currently, we use the depd and debug packages both for emitting deprecation warnings and for debug logging. I believe we should start using Node.js built-in utilities [util.debuglog](https://nodejs.org/api/util.html#utildebuglogsection-callback) and [util.deprecate](https://nodejs.org/api/util.html#utildeprecatefn-msg-code-options) for these purposes.

This would primarily reduce our dependency footprint and lower the risk of our packages being affected if those dependencies were ever compromised. In addition, while these packages are maintained by Doug, he has been inactive for some time.

As a result, if any package could previously be used in the browser, it will no longer work out of the box and users will need to bundle it themselves (e.g. using Webpack) to transpile it and make it browser-compatible, additionally, the way deprecation warnings are disabled will change to using Node.js options as described in the official documentation, which makes this a breaking change. Debug logging will also now be enabled via NODE_DEBUG instead of DEBUG.

This is not a major concern, as the only package we currently support for browser usage is iconv-lite, and it does not rely on either of these dependencies. None of the other packages claim browser compatibility.

jonchurch:
Replacing `depd` with builtin is low friction and has some benefits like only emitting once per deprecation code

replacing debug has low benefit and some tradeoffs, namely moving from `DEBUG=` to `NODE_DEBUG=`, losing colorized output, and nested namespacing.

webketje:
Something I ended up doing for an at-work library I maintain:

```ts
type GenericLogger = {
  info: (message: string | object | any, ...meta: any[]) => void
  debug: (message: string | object | any, ...meta: any[]) => void
  error: (message: string | object | any, ...meta: any[]) => void
  [key: string]: any
}
```
This is compatible with `express({ logger: debugDebugger })` (default), `express({ logger: console })` (native) , `express({ logger: debugLog('express') })` (NODE_DEBUG), `express({ logger: winstonLogger })` (winston).
It would *not* be semver-breaking because the default is kept, but in order to scope to log 'channels', the channels would need to be prepended to the message instead of managed by `debug`. To keep `DEBUG=express:*` feature parity, express would need to define its own log filtering regexp/ fn.

Also NODE_DEBUG prepends a number before every log & uppercases the 'channel'.

Delapouite:
[...]

PR #509, opened 2026-07-16
docs: add RFC for shared logging facade (`@expressjs/log`)

bjohansebas:
Okay, this is my main idea for keeping logging implementation-agnostic: https://github.com/expressjs/discussions/issues/459.

The core idea is that it doesn't depend on any logging library. By default, it has no logger configured. If you call `.setConsumer()`, you provide the logging implementation to use.

If `NODE_DEBUG=express` is set (see the RFC for the reasoning behind this), it will instead write to `stderr` using the current `debug`-style output as a migration path.

The goal is to avoid taking a dependency on any specific logger, at least initially. It also makes the package platform-agnostic, so it could work in browsers as well if needed—for example, if packages like `cookie` ever wanted to use it in the future.

I'd like to implement this package once we've finalized the design

### DIFF
--- a/docs/rfc/0001-expressjs-log.md
+++ b/docs/rfc/0001-expressjs-log.md
@@ -1,0 +1,304 @@
+# Adopt a shared logging facade (`@expressjs/log`) across the Express ecosystem
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


sheplu:
wip here - nothing fixed https://github.com/expressjs/log

